PR #3, opened 2015-10-26
Case-insensitive check for resource.exists() only when resource is case-sensitive

paulklinkenberg:
While working on my Azure resourceProvider, I noticed an extra file listing is done when calling fileExists() on a non-existing file. This is done based on the Operating system being unix/linux, where it should actually check if the resourceProvider itself is case-sensitive.

As a side note: my current resource has thousands of files stored directly in the root directory. This case-insensitive check (dir listing) takes quite a while to load; are you sure this is the right way to do it?


paulklinkenberg:
Awesome, glad to be able to help :)

PR #4, opened 2015-11-07
LDEV-573 - Updating text and links in the admin.

andrew-dixon:
Updating the text and links in the admin as described on ticket LDEV-573.

### DIFF
--- a/core/src/main/cfml/context/admin/overview.cfm
+++ b/core/src/main/cfml/context/admin/overview.cfm
@@ -440,7 +440,7 @@ Error Output --->
 
 					<!--- Jira --->
 					<h3>
-						<a href="https://luceeserver.atlassian.net/" target="_blank">#stText.Overview.issueTracker#</a>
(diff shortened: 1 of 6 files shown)


andrew-dixon:
My editor appears to have stripped a lot of whitespace from the edited files which makes the diff very difficult to see the changes.

modius:
Nice.

bdw429s:
Just add `?w=1` to the end of the URL to ignore whitespace.

https://github.com/lucee/Lucee/pull/4/files?w=1

andrew-dixon:
@bdw429s Good to know.

michaeloffner:
Why moving away from issues.lucee.org, I have chosen this URL so we get independent of the bug tracker we use. It is not impossible that in the future the URL of the bug tracker changes, when we use issues.lucee.org we only can change the direct and every update still points to the right point.

andrew-dixon:
Because the link doesn't work, but I think that is because it was https instead of http. I'll update it all later.

andrew-dixon:
@michaeloffner I've updated the issue tracker links as requested.

PR #6, opened 2015-12-09
fixes for LDEV-655 and building on Windows

dajester2013:
This fixes LDEV-655 by returning DIALECT_CFML from PageSourceImpl.getDialect() when the lucee dialect is disabled.

Also, there are two windows build fixes:
1. set the javac task's encoding to utf-8 (RHExtension.java would not compile)
2. an assertation in the resources test was coded with *nix file separators - changed to use platform-specific separator

### DIFF
--- a/ant/build-core.xml
+++ b/ant/build-core.xml
@@ -322,7 +322,7 @@ LUCEE BUILD PROCESS
       srcdir="${srcInst}" 
       target="1.7" 
       destdir="${temp}/agent"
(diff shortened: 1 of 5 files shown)


dajester2013:
So the test builds are failing due to being "unauthorized" in the deploy to sonatype.  All test cases are actually passing.

dajester2013:
closing this for a few reasons.
1. the utf-8 issue was solved by igal in a separate commit
2. this PR is cluttered with unrelated testcase fixes
3. i'm working on some other changes that make this unnecessary

PR #8, opened 2016-01-23
fixes https://luceeserver.atlassian.net/browse/LDEV-703

dajester2013:
Pass a new evaluator pool to the transform for loop condition.


micstriit:
this was already fixed with ticket 707, sorry was not aware of that pull request at the time we solved 707

PR #10, opened 2016-01-30
allow to disable controller via system property lucee.controller.disabled

isapir:


isapir:
added in a5eb9a8a2d3e019add45b6111ac3a63ba0eaed14

PR #16, opened 2016-03-17
Fix for JDBC drivers that do not implement Statement.getMoreResults(int) correctly

dajester2013:
Since fillResult closes the ResultSet already, just use Statement.getMoreResults() instead, as some JDBC drivers may not implement getMoreResults(int) correctly.


micstriit:
ATM i'm not sure what happens if the last object on the stack was an update count ... need to run some tests with different drivers...

PR #18, opened 2016-04-02
Styling of cfdump is broken in file output

paulklinkenberg:
Ref: https://groups.google.com/forum/#!topic/lucee/RTqGFb3ZEjg


isapir:
This is ok for a quick fix, but the styles should work with the class `.-lucee-dump` rather than the id.

Now that we have the improved htmlhead and htmlbody tags we can do it better.  I'll try to work on that sometime soon.

PR #19, opened 2016-04-07
Fix LDEV-812

jzuijlek:
Use complete filename in FileLogPool


micstriit:
we have adapted your fix, but we had to change it a little bit, see the commit link in the ticket for details

PR #20, opened 2016-04-08
Fixes: LDEV-574

wfears:
Using BigDecimal() setScale() to round number before passing it to NumberFormat.format()

@joel-ferreira was able to illustrate rounding errors with the following code snippet:

```
<cfoutput>
<cfloop from="1" to="999" index="i">
    <cfset number1 = 1 />
    <cfset number2 = number1 + (i/1000) />
    <cfset number3 = dollarFormat(number2) />
    <cfset number4 = '$' & numberFormat(number2, '0.00') />
    <cfif number3 NEQ number4>

    #number1#<br />
    #number2#<br />
    #number3#<br />
    #number4#<br />
        PROBLEM<br />
         <hr />
    </cfif>
</cfloop>
</cfoutput>
```

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/international/LSCurrencyFormat.java
+++ b/core/src/main/java/lucee/runtime/functions/international/LSCurrencyFormat.java
@@ -21,6 +21,7 @@
  */
 package lucee.runtime.functions.international;
 
+import java.math.BigDecimal;
 import java.text.NumberFormat;
 import java.util.Currency;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


micstriit:
i dont think that using BigDecimal is the right way, this slows down the function trastically

micstriit:
what does the trick in your change is this "BigDecimal.ROUND_HALF_UP"
but for that we need not to convert the double to a string, then to a BigDecimal and back to a double. that is much to "expensive" and not necessary. what we need is to round right.

micstriit:
we cannot pull this, because this breaks the function lsCurrencyFormat, for details please consult the ticket 574

PR #25, opened 2016-06-21
Lucee 5.0 thread stop issue

goochjj:
Use of thread.stop0 will only work on java.lang.Thread directly - fix reflection to use the parent class.

Specific cases of this - reported by @mjhagen on lucee channel on Slack... Related to database errors within a scheduled task.  Errors on the console are non-descript, and hide the actual error returned from the code.  stop0 isn't able to be executed on a TaskThread, so stop() is the only recourse, which sends ThreadDeath instead of the appropriate exception.


goochjj:
Note LDEV-893

Thanks!

PR #29, opened 2016-07-12
LDEV-919

NicholasClaaszen:
Add lucee- prefix to classes in dump to prevent shenanigans with CSS frameworks


NicholasClaaszen:
I sincerely hope I've done this correctly.

PR #36, opened 2016-08-30
Updated pom.xml to make project more generic and importable to other …

isapir:
…IDEs

Also added developer and reformatted xml so ignore whitespace to see changes better


isapir:
@michaeloffner Please review.  You can ignore whitespace by visiting https://github.com/lucee/Lucee/pull/36/files?w=1

The main thing IMO is that I changed the scope of dependencies from `provided` to `compile` so that they will be on the classpath at runtime, but it looks to me like the produced jar is the same.

isapir:
I left some environment specific absolute paths by mistake.  Will have to fix that.

PR #43, opened 2016-10-04
Potential fix for issue LDEV-1020

mattlevine:
I decided to take a stab at solving this issue and checking out the new Lucee 5 build process.  It super simple to contribute now!  I won't be offended if this is rejected. However, it's fairly important for me that Lucee support sending form params with PUT and PATCH http requests.  Otherwise I'll be stuck with a third party java lib which would seem dirty.

thanks,
-Matt


mattlevine:
Seems like my pull request picked up some extra cruft.

PR #44, opened 2016-10-04
Potential fix for issue LDEV-1020

mattlevine:
This is my attempt at fixing issue LDEV-1020.  

https://luceeserver.atlassian.net/browse/LDEV-1020

It would be great if Lucee could support sending form variables with PUT and PATCH http requests.  We need it to consume third party rest apis.  Otherwise we'll end up using a third party java lib for certain request which seems odd.

ACF already supports it.

Also, I love the 5.0 build process!

Thanks,
-Matt


micstriit:
i did adapt your changes with minor changes, thanks.

PR #46, opened 2016-10-13
fix for LDEV-811

nopjmp:
This checks for control characters and escapes them. I didn't write a test for this, but I manually tested and made sure it worked.

I did this against the 5.1 branch since it's the development branch currently


nopjmp:
did test the special controls we already take care off, making a new pull request with fixes

PR #47, opened 2016-10-13
fixes LDEV-811

nopjmp:
move the performance check down and pull out the switch. We will check our special characters first and then evaluate isISOControl to see if we need to escape it. We will then check for >= 128 before checking the encoder to avoid enduring the performance penalty of Charset.canEncode.

This is easier to understand from my last pull request. Additionally, I forgot that Carraige Return and Line Feed were control characters and were getting caught by the isISOControl test since I didn't great a test script.


micstriit:
thanks for your pull request. we will review the changes asap and add some testcases for it. Any input on this is of course welcome.

PR #51, opened 2016-10-27
Adjusted language in some down/up text for clarity

awayken:


micstriit:
im not sure about this change:
-stText.services.update.downUpTitle="Update / Downgrade";
+stText.services.update.downUpTitle="Update Lucee Version";

why remove "Downgrade" in this context?

awayken:
@micstriit In general, my update strives to clarify that this entire section of the admin is for updating your Lucee version.  Update is not the opposite of "downgrade." "Upgrade" is. Where there was the indication of downgrading, I updated the language to say "upgrade" to better reflect the proper parallelism.

In terms of the downUpTitle, I chose to reduce the redundancy of "Upgrade / Downgrade" for the more general (and equally clear) "Update Lucee Version." This falls better in line with the section above wherein one can change the properties of the Lucee update provider.

I hope this helps!

PR #52, opened 2016-10-27
solved https://luceeserver.atlassian.net/browse/LDEV-1040

isapir:
sort members of a Component's this scope when dumping


isapir:
@michaeloffner I added a small method to the Loader at https://github.com/lucee/Lucee/pull/52/files#diff-bb933c13de07d7c4e2d33fbac284725fR269

`rows` is a final variable so that was the cleanest way to handle this

micstriit:
doing any change to the loader has consequences and always makes troubles, because the core then depends on changes in the loader not everyone has.
Force people to manual update the jar makes no sense only for this.

so i would prefer to handle it as follows. Only make changes in the core and add some "FUTURE" marks to the loader to add this to the loader with Lucee 6.

isapir:
OK, I'll write it differently

PR #58, opened 2016-11-01
Added the test cases for LDEV-613

cfmitrah:


micstriit:
testcase must use function _internalRequest instead of cfhttp
https://github.com/lucee/Lucee/blob/master/test/general/Scope.cfc#L33

PR #67, opened 2016-11-07
Added test cases for LDEV-784

cfmitrah:


andrew-dixon:
@cfmitrah - I don't think these test cases are valid as it is checking the CSRF token in the same request as it is generated. You need to submit the CSRF token in a form and validate it on the next page and you will see the reported issue, which still exists in 5.2.2.71.

PR #74, opened 2016-11-08
https://luceeserver.atlassian.net/browse/LDEV-1068

Ronnie7:
Bug fix where SerializeJSON function is not returning uppercase column names inside "DATA" struct for Query Object.

### DIFF
--- a/core/src/main/java/lucee/runtime/converter/JSONConverter.java
+++ b/core/src/main/java/lucee/runtime/converter/JSONConverter.java
@@ -243,7 +243,6 @@ public final class JSONConverter extends ConverterSupport {
         	}
         }
     	
-    	
     	sb.append(goIn());
         sb.append("{");
         //Key[] keys = struct.keys();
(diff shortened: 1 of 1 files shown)


isapir:
You're on the right track, but we must first consult the "Dot Notation" setting and only call toUpperCase() if it is not set to "keep original case", otherwise you will fix your issue but introduce a similar issue to all of the other users who rely on the variables preserving the original case.

Ronnie7:
I was confused about "Dot Notation" behavior and turns out it has nothing to do with SerializeJSON function. SerializeJSON only returns a JSON string and it should return the same format as it is returning in Adobe Coldfusion.  
If you look at the JSONConverter.java:413 , you will see column names are uppercase without any "Dot Notation" check and it should be same in the '/DATA/' struct JSONConverter.java:426

Ronnie7:
@michaeloffner Whenever you have time, could you take a look at this pull request and merge it. It's a fix for [LDEV-1068](https://luceeserver.atlassian.net/browse/LDEV-1068)

micstriit:
@Ronnie7 i will asap, thanks for the pull

adamwinter2000:
@michaeloffner , any update on this?  This bug is preventing us from migrating an application from Adobe ColdFusion.  Thanks!

Ronnie7:
@michaeloffner Is there any chance to get this pull request done anytime soon? Once this is merged we can use Lucee snapshots for our application.

micstriit:
please see comments here
https://luceeserver.atlassian.net/browse/LDEV-1068

PR #79, opened 2016-11-10
Added test cases for LDEV-1068

cfmitrah:


test/tickets/LDEV1068.cfc
```diff
@@ -0,0 +1,18 @@
+component extends="org.lucee.cfml.test.LuceeTestCase"{
+	function run( testResults , testBox ) {
+		describe( "list to array function with numeric value", function() {
+			it("simple list to array function", function( currentSpec ){
```
awayken: The `describe()` and `it()` methods appear to be pasted from another test case. They should probably be rewritten for this test case. Maybe something like:

``` cfml
describe( "SerializeJSON", function() {
    it("should return uppercase column names inside DATA structure", function( currentSpec ){
```

awayken:
I think this pull request is great. The more tests, the merrier. However, I have a concern. There's a lot of manipulation happening after the `serializeJSON()` call. Is there a way to move the comparisons closer to that line of code? I'm just worried that this test case isn't isolating enough.

I wonder if doing simple `find()` calls for the column declarations as uppercase would suffice.  Something like:

``` js
    json = serializeJSON(query,true);
    expect(find('"NAME":', json) > 0).tobe(true);
    expect(find('"AGE":', json) > 0).tobe(true);
    expect(find('"SEX":', json) > 0).tobe(true);
```

cfmitrah:
@awayken Yes, it is a valid case. Actually we've thought something like

`expect(find('AGE', json) > 0).toBeTrue();`

But if a person named like **BOBBAGE** then case will be broken. That why we just make it by getting column list & then compare it with original columns list.

But with 
`find('"AGE":') > 0`
The issue won't happens, so applied your suggestion in new commit. As well as i've made some changes too...

awayken:
@cfmitrah This looks great. I think this version of the test most accurately reflects the issue. Thanks for your work on this!

PR #90, opened 2016-11-15
Added test cases for LDEV-378

cfmitrah:


micstriit:
this is not a bug, it is expected behaviour.

PR #114, opened 2016-11-28
BugFix: No matching Method/Function for Query.getColumnList(boolean) found.

Ronnie7:
@michaeloffner Thank you for making update for  [LDEV-1068](https://luceeserver.atlassian.net/browse/LDEV-1068) we really appreciate it. Please take a look at this pull request. You may see lots of white spaces changes please ignore them it is because of my editor using 4 tab formatting.
**Details:**
Changing the argument type from boolean to PageContext in the getColumnList method causes error  `No matching Method/Function for Query.getColumnList(boolean) found  `.  I fixed this issue by  creating a private method called `getKeyCase( PageContext pc )` which returns boolean keycase value of true or false.
Pass this value to `getColumnList` method, which will take care of upper/lower case without changing the method definition.


Ronnie7:
@michaeloffner Have you got a chance to look at this pull request. This is a hot fix for bug got introduced in this this ticket ([LDEV-1068](https://luceeserver.atlassian.net/browse/LDEV-1068)).

Ronnie7:
@micstriit Any update on this pull request.

micstriit:
sorry the delay, we can do this, also in branch 5.1 what is the branch where bug fixing is happening.
But i think we should not get rid of getColumnList(pc). But we have to make sure this method is not used in any extension of us.
Maybe best keep both methods...

micstriit:
can you raise a new ticket for this issue?

Ronnie7:
@micstriit This error was introduced with the 5.1.1.30 snapshot.

micstriit:
please add that info to the ticket (affected version 5.1.1.30 snapshot), i did look into it. It is ok to remove getColumnlist(pc) because it is used nowehere.
But there is an other problem with the pull request, the method keyCase(...) is called even it is not used. it should only be called if getColumnList is called afterwards. in addition it is not necessary to store the result in a variable, it is always only used once as far as i know. all that makes the execution of the code slower.

Ronnie7:
@micstriit  I created a JIRA ticket for this issue here is link [[https://luceeserver.atlassian.net/browse/LDEV-1160](https://luceeserver.atlassian.net/browse/LDEV-1160)]

micstriit:
instead of this:
```
boolean upperCase = getKeyCase(ThreadLocalPageContext.get());
...
if(key.equals(KeyConstants._COLUMNLIST)) return getColumnlist(upperCase);
```
you can simply do this:
```
if(key.equals(KeyConstants._COLUMNLIST)) return getColumnlist(getKeyCase(ThreadLocalPageContext.get()));
```

Ronnie7:
@micstriit I did it for readability purpose. however, if it's going to slow down code execution I better update it as you suggested. Thank you for suggestion 👍

micstriit:
have in mind that a method like [this](https://github.com/lucee/Lucee/blob/5.2/core/src/main/java/lucee/runtime/type/QueryImpl.java#L832)  can be called 1000 of times when looping just one query. that is btw the reason that this method had the check on line [837](https://github.com/lucee/Lucee/blob/5.2/core/src/main/java/lucee/runtime/type/QueryImpl.java#L837). even we added a addional step, is is faster in the end.

Ronnie7:
@micstriit I have updated the file as you suggested. Please take a look and let me know If I need to make any other changes. Thank you.

PR #120, opened 2016-12-01
Additional logging in scheduled tasks

cubiclabs:
Additional logging was added to scheduled tasks while looking into problems detailed in this ticket:
https://luceeserver.atlassian.net/browse/LDEV-897

I thought that it might be useful going forward.

### DIFF
--- a/core/src/main/java/lucee/runtime/schedule/ScheduledTaskThread.java
+++ b/core/src/main/java/lucee/runtime/schedule/ScheduledTaskThread.java
@@ -88,6 +88,10 @@ public class ScheduledTaskThread extends Thread {
 		try{
 		_run();
 		}
+		catch (Throwable t) {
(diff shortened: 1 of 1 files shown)


yamikuronue:
Please merge this PR. It would have saved us days of frustration at my work if we had some way to see why tasks were silently failing while we were debugging them.

isapir:
Is there a reason you used DateTimeImpl instead of the CFML functions like DateFormat() and DateDiff()?

cubiclabs:
I just tried to copy what was already there. DateTimeImpl was already being used in log output so assumed that was the right thing to use.

My knowledge of Java is minimal.

cubiclabs:
Is there anything else I need to do in order to get this accepted??

isapir:
Sorry for the delay.  I will try to look at this today.

PR #186, opened 2017-03-14
Reimplementing fix for LDEV-364

SimonHooker:
The fix for LDEV-364 appears to have been lost during migration of LDEV-224.  This reimplements that fix to allow null= to be used once again.


SimonHooker:
The failures within Travis are relating to the S3 implementations - untouched by this commit

PR #211, opened 2017-04-19
Added test case for LDEV-1197

cfmitrah:


micstriit:
added manually to branch 5.2

PR #224, opened 2017-05-03
Added test case for LDEV-1292.

cfmitrah:


test/tickets/LDEV1292.cfc
```diff
@@ -0,0 +1,106 @@
+component extends="org.lucee.cfml.test.LuceeTestCase"{
+	// skip closure
+	function isNotSupported() {
+		var mySql = getCredentials();
+		if(!isNull(mysql)){
```
duchuyctlk, line 5: why don't you just return `isNull(mysql)` ?

micstriit: we will look into it when we address the ticket, thank for the input.

PR #235, opened 2017-05-15
Fixed the issue on CFDump when preserve case is enabled.

cfmitrah:


mattiLeBlanc:
When will this be available in Lucee 5.2.x? I just upgraded a new server to latest Lucee and can't see my error message because I get the same 
```
(index):22 Uncaught TypeError: Cannot read property 'TITLE' of undefined
    at parseDumpJSON ((index):22)
    at (index):127
```
message.
I am running Lucee 5.2.19 on Ubuntu 17

mattiLeBlanc:
I can verify that a lot of the JS code generated is using Uppercase property references:
```
if(currDumpObj.HEIGHT!="")
tbl.style.height=currDumpObj.HEIGHT;
if(currDumpObj.TITLE!="")
tbl.title=currDumpObj.TITLE;
if(typeof(currDumpObj.META.TITLE)!="undefined"){
var titleTR=document.createElement("tr");
var titleTD=document.createElement("td");
if(typeof(currDumpObj.META)!="undefined"){
if(typeof(currDumpObj.META.ONCLICK) != "undefined"){

```
while the code clearly provides a dump with lowercase `meta.title` etc. Please fix asap.

When I disable 'preserve case' it works however when I get an error and view that in my chrome dev tools in the preview tab, the JS is not executed and thus the error is not rendered. 
I have to copy the source into a local HTML file and view locally.
Is this expected behaviour?

cfmitrah:
@mattiLeBlanc We've fixed this issue already in https://github.com/lucee/Lucee/pull/236 Actually we've made the original dump back for default cfdump & added a new format "javascript" for the new implementation.

mattiLeBlanc:
Is it already available as a patch? I installed the latest version 3 days ago and that still have the issue (Lucee 5.2.1.9)

mattiLeBlanc:
@cfmitrah I have no clue how to fix this issue. My version 5.2.1.9 does not have the patch and I can't find an option to set cfdump back to default behaviour. Yes, I do use PRESERVE CASE, but should just work.
When is a patch available?

mattiLeBlanc:
Even when I enable cfdump preserve case and do a `cfdump` in my cfm code I still get a JS error
```
(index):294 Uncaught TypeError: Cannot read property 'TITLE' of undefined
    at parseDumpJSON ((index):294)
    at (index):399
```

PR #250, opened 2017-06-20
Ldev 238

cfmitrah:


isapir:
merging into branch LDEV-238 for review

isapir:
@cfmitrah here are a few issues that I see:

The condition in line 2 is the same as in line 5 at https://github.com/lucee/Lucee/pull/250/files#diff-f25ce9af872284dba3f27e92b48a5185R2 ?  If so, then the two `cfif` checks should be combined.

The regex `^\d+(\.\d+)*$` does not account for IPv6 at https://github.com/lucee/Lucee/pull/250/files#diff-2b3f668bd0f52b608ddb4f475d5d45a5R4

The Java access `<cfset filePath = getPageContext().getConfig().getConfigDir().getRealResource("lucee-web.xml.cfm")>` will throw an error if Java access is disabled in the security settings at https://github.com/lucee/Lucee/pull/250/files#diff-968144167e95092bfb0618341610cc9bR37

What is the IP `127.0.0.12` at https://github.com/lucee/Lucee/pull/250/files#diff-421d853e1e2e7ebfc232111fecad9598R434 and https://github.com/lucee/Lucee/pull/250/files#diff-cbbf51ceff2d128a7d13a9e654446df4R236 ?  Looks invalid to me.

On some operating systems, like Windows, the "outgoing" IP is not necessarily `127.0.0.1` or `::1`.  For that reason we added `GetLocalHostIP(true)` which returns an array of all of the local IP addresses, as I mentioned in the ticket.  We should have a "checkbox" for "Allow Local IPs" or something like that, which when checked, will check that `GetLocalHostIP(true)` contains `CGI.REMOTE_ADDR`.  Additional IPs should be stored in the XML.

PR #252, opened 2017-06-21
LDEV-1386 for Master

jzuijlek:


micstriit:
we never to changes on the master branch directly, the only way the master branch get changed is by merging current state of the stable branch (5.2 atm) into the master branch.

PR #255, opened 2017-06-26
LDEV-357

SimonHooker:
Adding flag to use RSET instead of NOOP to improve bulk mail performance ( https://luceeserver.atlassian.net/browse/LDEV-357 )


isapir (commented):
@SimonHooker I imagine that you tested that?  I don't see an issue with it, but can you please open a JIRA ticket for this as well and describe whatever issues you were seeing before this?

The docs say ` In some cases sendmail will respond slowly after many NOOP commands; use of RSET avoids this sendmail issue.` which seems fine to me.

PR #260, opened 2017-07-12
Fixed the error in mail server verification issue.

cfmitrah:


mabusalah:
This did not solve the error

PR #261, opened 2017-07-12
Added checkbox for allowing local IP's

cfmitrah:


isapir:
@cfmitrah When @micstriit said to adapt the debugging IP solution he meant to "model after that", not to change it in the Debug section.  I believe there was a miscommunication there.  The solution here is to protect the Admin, so it should not be mixed with the Debugging.  A more appropriate place would be in `security.access`.

I got different errors when I tried to run your code.  For example, on my existing Lucee deployment, I have no entries for `ipresrtriction`, so 

```XmlSearch(xmlObj, '//*[ local-name()=''ipresrtriction'' ]')``` 

always returns an empty array, and attempting to access an element, e.g. 

```deserializeJSON(xmlElem[1].xmlText``` 

throws an error.  This happens in several places due to code duplication.  There should be one function that checks for it and return valid defaults if the configuration does not exist.

TBH This shouldn't happen at all this way.  We should add the functionality to `cfadmin`, and the Admin template code should make the call to `cfadmin` just like the rest of the configuration values.

Another error that I encountered was 

>The key [ADDIP] does not exist, only the following keys are available: [ALLOWIPDESC].

at debugging.templates.ipedit.cfm: line 14 

I think that we should discuss this before the next attempt.

PR #264, opened 2017-07-25
Added stack trace in Collapsible div mode

cfmitrah:


isapir (changes requested):
We already use jQuery in the Admin, so better to use known patterns.  See for example how we do that at https://github.com/lucee/Lucee/blob/5.2/core/src/main/cfml/context/admin/admin_layout.cfm#L124

Please set all JS code to render at the bottom of the page.  You can do that by using `<cfhtmlbody>`.  See example at https://github.com/lucee/Lucee/blob/5.2/core/src/main/cfml/context/admin/debugging.settings.cfm#L97

Please use camelCase for JS variables and function names, and kebab-case for CSS id and classes.

Thanks :)

PR #284, opened 2017-09-06
Added thread for extension update

cfmitrah:


micstriit:
we need to make this in a different way

PR #289, opened 2017-09-11
Remove ajax factory

cfmitrah:


micstriit:
first the ajax extension need to be in place befirte we can remove this

PR #298, opened 2017-09-15
Hot Fix

cfmitrah:


micstriit:
i had to rewritte a lot, so i cannot integrate this

PR #305, opened 2017-09-23
Collapse SQL results in modern debugging output

zspitzer:
- combine sql queries into a single table
- add toggle to display sql source
- highlight slow queries, expand slow queries by default, hide sql for all other queries
- add a couple of missing var statements

https://luceeserver.atlassian.net/browse/LDEV-1515


isapir:
@zspitzer Thank you.  A couple of points:

>add a couple of missing var statements

This can go in `5.2` but the rest should go to `5.3`.  for the sake of efficiency, the whole thing should probably go to `5.3` though.

>add toggle to display sql source
hide sql for all other queries

Are the toggle settings "remembered"?  If not, it can be annoying to keep expanding them for someone who wants them always opened.

PR #310, opened 2017-09-25
LDEV-1515 Collapse SQL results in modern debugging output

zspitzer:
combine sql queries into a single table
add toggle to display sql source
highlight slow queries, expand slow queries by default, hide sql for all other queries
add a couple of missing var statements

BTW I'd love to add query resultset size to the debug output, it would be really useful for monitoring memory usage

### DIFF
--- a/core/src/main/java/resource/context/admin/debug/Debug.cfc
+++ b/core/src/main/java/resource/context/admin/debug/Debug.cfc
@@ -52,6 +52,17 @@
     	<cfthrow message="implement function output():string">
     </cffunction>
     
-    
+    <cfscript>      
+        variables.serverTimingHeaders = [];
+
+        function addServerTimingHeader(metric, name, timeMs){
+            arrayAppend(variables.serverTimingHeaders, '#arguments.metric#=#arguments.timeMs#; "#arguments.name#"');
+        }
+
+        function outputServerTimingHeaders(){
+            if (not getPageContext().getHttpServletResponse().isCommitted() ) // avoid cfflush error
+                header name="Server-Timing" value="#ArrayToList(variables.serverTimingHeaders,", ")#";
+        }
+    </cfscript>    
     
 </cfcomponent>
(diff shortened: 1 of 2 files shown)


zspitzer:
regarding the earlier question about persisting expanded state, there isn't enough information to identify individual queries and the reference would need to include the page url, so even if we could persist a reliable reference, it would potentially be a rather large cookie

isapir (commented):
What's up with the empty anchor tag?
`<a>#queries.name#</a>`

If we can't save the toggle the state then IMO we should not collapse certain queries by default.

Did you try to add a Memory size of the Queries similar to the way we do for the Scopes?  This is how we show the _estimated size_ of the Scopes:
https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/resource/context/admin/debug/Modern.cfc#L743

zspitzer:
the empty anchor tag on the query name is to provide a visual indicator that you can click to expand the results to see the sql, but the actual onclick handler is for the whole row.

the query object doesn't include a reference to the actual query, it only contains the following columns,
name | time | sql | src | count | datasource | usage | cacheType

Something needs to be done in java to add size info

Regarding persistence, how about I just add a persistent expand all sql cookie?

isapir:
>Regarding persistence, how about I just add a persistent expand all sql cookie?

I really don't see the benefit in that since the whole section can be toggled.  I think that adding some visual signal for the slow queries is a good idea.  Only if you have many queries in the Request does it make sense to go through the trouble of collapsing/expanding.

isapir:
>the query object doesn't include a reference to the actual query, it only contains the following columns,
name | time | sql | src | count | datasource | usage | cacheType

You can open a new ticket requesting to add the Query memory footprint to the debug template.

zspitzer:
done! https://luceeserver.atlassian.net/browse/LDEV-1517

zspitzer:
I have added a persistent toggle to expand all sql queries

Even if you only have a few queries in a request, the sql can be rather long and span many pages,
this makes the debug output much more concise by default, whilst still preserving the old verbose behavior

zspitzer:
I also have an implementation of Server timing headers which i could add to this pull request 
https://luceeserver.atlassian.net/browse/LDEV-1520

zspitzer:
ok, i'll submit a new PR against 5.3

cfmitrah:
Hi @zspitzer,

I've added server timings  into latest 5.3 branch. Can you please check it

Pull Request: https://github.com/lucee/Lucee/pull/468

PR #312, opened 2017-09-25
Modified the extension page

cfmitrah:


micstriit:
@cfmitrah still relevant?

cfmitrah:
@micstriit , I think  this changes is not needed in 5.3. we changed the code already. so I close this Pull Request

PR #313, opened 2017-09-25
Added code to load the plugin without restart lucee

cfmitrah:


micstriit:
i don't think we should make a directory listing with every single request.
instead of the QoQ you can use the function "QuerySort"

PR #317, opened 2017-10-01
LDEV-1530 - Admin show bundle last updated 

zspitzer:
show when a bundle was last updated in admin

image

https://luceeserver.atlassian.net/browse/LDEV-1530

### DIFF
--- a/core/src/main/cfml/context/admin/info.bundle.list.cfm
+++ b/core/src/main/cfml/context/admin/info.bundle.list.cfm
@@ -3,6 +3,20 @@
 	password="#session["password"&request.adminType]#" 
 	action="getBundles" 
 	returnvariable="bundles">
+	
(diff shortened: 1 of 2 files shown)


micstriit:
i have 175 bundles in my regular installation, that means the function getBundleLastModified is called 175 times. means we list 175 zip files and make 175 QoQ.
That is in my opinion a bridge to far.

adding this functionality to the detail view, where getBundleLastModified is called only once is fine to me.

micstriit:
Better is to use "Bnd-LastModified: 1384987499624" from Manifest, what we load anyway

micstriit:
https://luceeserver.atlassian.net/browse/LDEV-1816

extracted build date from manifest that is already loaded anyway (bundle.headers), not all have a date in the manifest but a lot do.

PR #355, opened 2017-12-27
LDEV-339 : arrayContainsNoCase looks at item substring

cfmitrah:
I've added test cases for this ticket & fixed the issue by changing the ArrayContainsNoCase.java a bit.


micstriit:
manually adapted change because there are conflicts with current version

PR #364, opened 2018-01-08
Update Input.java

CraicOverflow89:
It was throwing when seven day names were supplied rather than when there weren't seven. Also changed the @param to correct argument name.


micstriit:
this tag is only in the core as hidden feature because the Lucee admin make use of it. The form tags itself have moved to an extension, becausde of that we have adapted this fix in the extension.

PR #390, opened 2018-02-12
Modified the test case for LDEV-1621

cfmitrah:


micstriit:
adapted change manually because of conflict

micstriit:
https://github.com/lucee/Lucee/commit/dd5ded17f699f9fed549676958a057939e4b5b99

PR #405, opened 2018-03-02
Added QueryBuilder component

cfmitrah:


micstriit:
we will do this in a extension

PR #411, opened 2018-03-09
Added test case for LDEV-1718

cfmitrah:


micstriit:
this testcase only makes sense in case it runs on a windows system and drive "d" is present

bdw429s:
You can change it to be any drive letter you want.  In my case, it just happened to be a D drive.

micstriit:
we would need to have a windows server to run this testcase, what we do not have and not plan to have in the near future

PR #413, opened 2018-03-13
Added count attribute in listLast function

cfmitrah:


micstriit:
this code does not handle includeEmptyField at all

PR #415, opened 2018-03-14
Added collection component for 5.3

cfmitrah:


micstriit:
this need to be added to the Lucene Extension, the search is not part of the core

PR #440, opened 2018-04-05
performance.cfc focussed debug template

zspitzer:
modified modern template, focussed on highlighting performance issues
- execution times are sorted slowest to fastest
- slow sql queries are collapsed by default
- server timing headers, performance stats show up in dev tools under timing

replaces https://github.com/lucee/Lucee/pull/310 which modified the modern.cfc


micstriit:
Actually we don't want to add a new Debugger to the core, specially not to the 5.2 branch that only gets bug fixes and not new features.
We actually even consider removing the "Comment" debugger from the core and move to an extension.

Sure we are open for new Debugger outputs and/or improving the existing ones.

Improving the existing "Modern" debugger need to happen on the "5.3" branch, because it is completely rewritten in 5.3 (added tabs for monitor and doc). 

A new "Performance" debugger need to happen as a Extension, i will see if we can move the "Comment" debugger to an extension that the can be used as a template.

zspitzer:
ok.

PR #441, opened 2018-04-06
LDEV-1630 and LDEV-1757

bdw429s:


bdw429s:
@michaeloffner @micstriit

PR #442, opened 2018-04-09
Trim too long string

cfmitrah:


zspitzer:
can the string be escaped to avoid xss?

micstriit:
did manually adapt the solution
https://github.com/lucee/Lucee/commit/6f01a72a4617894088d9054c6f3ad13c46878518

this code has 3 problems
1. it use StringUtils from apache, what we never do
2. it has unnecessary redundant code
3. it does not cover all occurrences of this case

PR #450, opened 2018-04-16
add cfzip compress option LDEV-930

zspitzer:
initial cut, not working yet, I need some java help!
https://luceeserver.atlassian.net/browse/LDEV-930

### DIFF
--- a/core/src/main/java/lucee/runtime/tag/Zip.java
+++ b/core/src/main/java/lucee/runtime/tag/Zip.java
@@ -64,6 +64,10 @@ public final class Zip extends BodyTagImpl {
 
 	private String action = "zip";
 	private String charset;
+	private LinkedList<ResourceFilter> compressor;
+	private ResourceFilter compress=null;	
+	private String compressPattern;
+	private String compressDelimiters;
 	private Resource destination;
 	private LinkedList<String> entryPathList;
 	private String[] entryPaths;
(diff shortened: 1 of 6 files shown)


isapir:
@zspitzer Thank you for the PR.  We really _do_ appreciate your contributions to the Lucee project.  This PR, however, suffers from several issues:

1) It introduces a new API that was not discussed/approved, by adding a new attribute named `compress`.  This may, or may not, be the right way to go, but it should be discussed and agreed upon first.

2) The code that you submitted does not compile.  It is time consuming to review the code and figure out what another developer had in mind in the first place, but even more so when the code does not compile.  

3) I modified the code and got it to compile in a mocked up test environment (read: not production ready), but then while testing the STORED method I got a runtime error: "STORED entry missing size, compressed size, or crc-32" at java.util.zip.ZipOutputStream.putNextEntry(ZipOutputStream.java:224) 

See https://stackoverflow.com/questions/1206970/how-to-create-uncompressed-zip-archive-in-java for more on that.

isapir (commented):
@zspitzer Thank you for the PR. We really do appreciate your contributions to the Lucee project. This PR, however, suffers from several issues:

1) It introduces a new API that was not discussed/approved, by adding a new attribute named compress. This may, or may not, be the right way to go, but it should be discussed and agreed upon first.

2) The code that you submitted does not compile. It is time consuming to review the code and figure out what another developer had in mind in the first place, but even more so when the code does not compile.

3) I modified the code and got it to compile in a mocked up test environment (read: not production ready), but then while testing the STORED method I got a runtime error: "STORED entry missing size, compressed size, or crc-32" at java.util.zip.ZipOutputStream.putNextEntry(ZipOutputStream.java:224)

See https://stackoverflow.com/questions/1206970/how-to-create-uncompressed-zip-archive-in-java for more on that.

PR #451, opened 2018-04-17
Conditionally skip a test that sometimes fails when system clock is not UTC.

sbleon:
This test fails when the local clock is in the America/New_York timezone and it is morning. `createODBCdate()` seems to treat all input as local time, and is not configurable. I think our only option is to skip the test under these circumstances.

### DIFF
--- a/test/jira/Jira2880.cfc
+++ b/test/jira/Jira2880.cfc
@@ -16,8 +16,11 @@
  * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
  * 
  ---><cfscript>
-component extends="org.lucee.cfml.test.LuceeTestCase"	{
-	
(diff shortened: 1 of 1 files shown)


micstriit:
to be honest i'm not sure if skipping the testcase is the right approach, i try never to suppress testcase. is there a reason we cannot improve the testcase, is this a bug?

micstriit:
https://github.com/lucee/Lucee/commit/0cac7ea1d2f7090f701518def8b7ef54bcb1f208

sbleon:
Cool, thanks! I'll have to try running the test suite in the morning
tomorrow to make sure it passes :-)

sbleon:
Looks good here in EST. Thanks, Michael!

PR #452, opened 2018-04-17
Added some changes in fld

cfmitrah:


micstriit:
this is based on the ACF 2016 doc and most likely not represent the actual names used for named arguments in ACF 2018. specially names like "time-string" will not work anyway, then doing
`func(time-string:"");`
will throw a Template (syntax) exception.

Also this pull request should be for 5.3 and not 5.2.

There is already an update in Lucee 5.3 based on this list
http://cfdownload.adobe.com/pub/adobe/coldfusion/2018/publicBeta/NamedParametersColdFusion2018.pdf

Still not sure if that pdf above will match the final release a 100%.
In the end we need to test against the final release of ACF 2018

I will keep this pull open for the moment.

micstriit:
sorry i need to close this, the choosen names are not accurate and we already have that change.

PR #468, opened 2018-04-23
 Added server time in 5.3 latest

cfmitrah:


core/src/main/java/resource/context/admin/debug/Debug.cfc
```diff
@@ -51,7 +51,17 @@
...
     </cffunction>
-    
-    
-    
+
+	<cfscript>
+		variables.serverTimingHeaders = [];
+
+		function addServerTimingHeader(metric, name, timeMs){
+			arrayAppend(variables.serverTimingHeaders, '"#arguments.name#" - #arguments.metric#=#arguments.timeMs#');
```
zspitzer: have you tried this out in chrome to see if it works? 

the format for the server timing headers has changed since my original commit
https://github.com/zspitzer/lucee-loganalyzer/blob/refactoring/source/cfml/plugins/renderUtils.cfc#L54

zspitzer: you can see these in the timing panel when you open a request under network in chrome dev tools

cfmitrah: @zspitzer, I'll check it & Let You Know

cfmitrah:
@zspitzer now I saw server timings in chrome dev tool

micstriit:
There is one problem with this code that need to be addressed and one concern (see tickets for detail).
you cannot do this code in CFML "getPageContext().getHttpServletResponse().isCommitted()", if "direct java access" is disabled in the Security/access Page in the server admin, you will get an exception with that code.
This template should not be affected by different admin settings.

The solution is not to test and simply catch the exception if they occur OR add a function to lucee (isFlushed) that gives this info.

zspitzer:
IsFlushed would be useful, it should be available.

Is there something like the old asp isConnected() available?

micstriit:
@zspitzer no there is no isConnected, but i like the idea, i will add both if possible, i'm not sure if the servlet engine is providing that kind of info.
I know that not all servlet engine are handling this in the same way. Some are stopping the request when the client disconnect, some don't.

micstriit:
every function we add brings the risk to break code in case user have the same function already in use. So instead of having 2 functions i prefer to have only one `GetContextInfo():{flushed:...,connected:...}`

micstriit:
added getContextInfo https://luceeserver.atlassian.net/browse/LDEV-1866 (ATM only provide flushed, more to come), can you adapt this function?

zspitzer:
I've been following the Firefox implementation of this, they mentioned Trailer headers

https://bugzilla.mozilla.org/show_bug.cgi?id=1403051
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

michaeloffner:
@zspitzer what is the status on this?

core/src/main/java/resource/context/admin/debug/Debug.cfc
michaeloffner: @cfmitrah is that still ready to go for 5.3? if so can you please redirect it to 6.0. i do not wanna make that kind of change to 5.3

michaeloffner:
this is very old if this still makes sense @cfmitrah @zspitzer please make a PR for Lucee 6.1

PR #475, opened 2018-04-27
expose services.restart in web administrator

zspitzer:
LDEV-1821


cfmitrah:
@zspitzer need to change on some Java files, PR is fine for CFML side

micstriit:
in my opinion this should not work.
Lucee only can restart the complete Lucee Server, not a single context.
So that action should 
a) do nothing
b) throw an exception

cfmitrah:
Hi @micstriit,

While restarting via web admin it doesn't do any thing. If we use code like below
`<cfadmin
    action="restart"
    type="web"
    password="password">`

It will throw error. 

Ticket :  https://luceeserver.atlassian.net/browse/LDEV-1839

micstriit:
as @cfmitrah pointed out, we cannot simply enable this, because it is not working YET

PR #494, opened 2018-05-17
Added test case for LDEV-1856

cfmitrah:


micstriit:
that testcase will not work in travis, travis runs Lucee via JSR223, so there is no webserver you can call.

PR #497, opened 2018-05-21
Fixes LDEV-398

Yamaha32088:


cfmitrah:
Hi @Yamaha32088 ,

Good work.., I've checked with above code working fine for me. Test case(LDEV398)  passes fine for me. @micstriit will confirm this ASAP.

micstriit:
I'm not sure this is the right way to go, enable this in general may cause problem where this is not expected, instead we need to make sure serializeJson is calling the "getProperties(b,b,b,b)" directly

Yamaha32088:
@micstriit ok I will try to make the changes tonight if I have time unless someone beats me to it.

micstriit:
@Yamaha32088 i was faster ;-)
https://github.com/lucee/Lucee/commit/7fed31554aed45fe9bfc9cd7e35ce4b21efaa266

PR #499, opened 2018-05-22
Fix administrator.cfc updateMapping()

jordanclark:
Code references missing "existing" variable which should pre-load mapping sttings.

### DIFF
--- a/core/src/main/java/resource/component/org/lucee/cfml/Administrator.cfc
+++ b/core/src/main/java/resource/component/org/lucee/cfml/Administrator.cfc
@@ -650,7 +650,21 @@ component {
 	* @inspect type of inspection for the mapping(never/once/always/"").
 	*/
 	public void function updateMapping(required string virtual, string physical="", string archive="", string primary="", string inspect="") {
-		var existing = getMapping(arguments.virtual);
+		var mappings = getMappings();
+		query name="exists" dbtype="query" {
+			echo("SELECT * FROM mappings WHERE virtual = '#arguments.virtual#'")
+		}
+		if( exists.recordCount ) {
+			if( isNull(arguments.archive) || isEmpty(arguments.archive) ) {
+				arguments.archive = exists.archive;
+			}
+			if( isNull(arguments.primary) || isEmpty(arguments.primary) ) {
+				arguments.primary = exists.primary;
+			}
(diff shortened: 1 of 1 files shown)


zspitzer (commented):
as isEmpty(NullValue()) returns true, is the isNull check required?

https://trycf.com/gist/4a185c5bb28ef5da002443f936d72202/lucee5?theme=monokai

micstriit:
we should replace QoQ with other code, Lucee only supports QoQ for compatibility reasons

zspitzer:
@micstriit  i can't edit the PR, but here is a replacement for the QoQ

```
var mappings = getMappings();
var _virtual = arguments.virtual; // avoid scoping conflict in closure
exists = mappings.some(function(row){ return row.virtual eq _.virtual});
if (exists)
```

dswitzer:
Here's the fixed method I'm using:

```
	// we need to overwrite the default handler, due to bugs in the code
	public void function updateMapping(required string virtual, string physical="", string archive, string primary, string inspect){

		/*
			We want to check if we're trying to update an existing mapping. If so, then we will use 
			the existing values for defaults for any arguments not supplied.

			NOTE - If an argument exists, but is blank, then we assume the user is trying to overwrite/clear
			the value. This is why we only check for unassigned values.
		*/
		try {
			// check for the existing mapping
			var existing = getMapping(arguments.virtual);

			// if values were not supplied, use the previous values
			if( isNull(arguments.archive) ){
				arguments.archive = existing.archive;
			}

			if( isNull(arguments.primary) ){
				arguments.primary = existing.primary;
			}

			if( isNull(arguments.inspect) ){
				arguments.inspect = existing.inspect;
			}
		} catch (Any e){
			// do nothing, because we had no existing value
		}

		admin
			action="updateMapping"
			type="#variables.type#"
			password="#variables.password#"

			virtual=arguments.virtual
			physical=arguments.physical
			archive=arguments.archive ?: ""
			primary=arguments.primary ?: ""
			inspect=arguments.inspect ?: ""
			toplevel="yes"
			remoteClients="#variables.remoteClients#"
		;
	};
```

zspitzer:
@dswitzer can you file a new PR with your approach, this PR is against 5.2.7, it needs to be against 5.3

couldn't find a bug, so i filed https://luceeserver.atlassian.net/browse/LDEV-2977

PR #501, opened 2018-05-25
fixes LDEV-929

Yamaha32088:


Yamaha32088:
@micstriit could you please take a look at this PR. I wasn't able to figure out how to re-trigger the build without making a pointless commit but I believe it passes all tests and something caused some environment variables not to be set which caused the tests to fail.

image

micstriit:
problem is, pull request have no access to env variables and this testcase not properly test if they are available and does not skip if not. i will fix

PR #509, opened 2018-05-31
Removes chart process cfc file in debug folder

cfmitrah:


micstriit:
Problem with that, the code does only avoid adding ChartProcess, but it does not remove existing one

PR #511, opened 2018-06-05
Changed error message for LDEV-1875

cfmitrah:


micstriit:
i adapted your change but in a little different way

micstriit:
https://github.com/lucee/Lucee/commit/251e9123618ebcbe6ba07e19a0653106f0b133ad

PR #516, opened 2018-06-09
Changes in debug output

cfmitrah:
1) Call chart process.cfc only when metric tab is visible
2) Changed the chart frequency
3) Changed all, javascript varables under __Lucee
4) Javascript files were loaded from external file instead of inline


micstriit:
please change "lucee-modern-extended" to "lucee-modern". changing the name causes problem for lucee instances updating

isapir:
Why not give it a new name and keep the old one (modem) as a separate
template?

micstriit:
@isapir because the template is not a new template called ModernExtended.cfc, it replaces the current Modern.cfc. in fact the id of the template should be ignored and instead the hash(filename) should be used as id. But atm that is not the case so we cannot change the id.

Having 2 "modern" template makes no sense. if we want to provide the old "Modern" template as well. we can provide it as an extension.

cfmitrah:
Closed the Pull due to conflicts with base branch 5.3

PR #525, opened 2018-06-20
Remove the echo in dump

cfmitrah:


cfmitrah:
Closed due to conflicts

PR #535, opened 2018-07-06
Correct spelling of daemon. 

sbleon:
"deamon" is a misspelling, and should not be considered valid.

### DIFF
--- a/core/src/main/java/lucee/runtime/tag/ThreadTag.java
+++ b/core/src/main/java/lucee/runtime/tag/ThreadTag.java
@@ -182,11 +182,15 @@ public final class ThreadTag extends BodyTagImpl implements DynamicAttributes {
 			//throw new ApplicationException("invalid value ["+strType+"] for attribute type","task is not supported at the moment");
 			type=TYPE_TASK;
 		}
-		else if("daemon".equals(strType)) type=TYPE_DAEMON;
-		else throw new ApplicationException("invalid value ["+strType+"] for attribute type","values for attribute type are:task,daemon (default)");
-		
+		// FUTURE remove "deamon", which is a misspelling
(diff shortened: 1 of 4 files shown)


isapir (changes requested):
@sbleon I agree with you, but unfortunately "APIs are forever", or at least for a very long time, and the change in core/src/main/java/lucee/runtime/tag/ThreadTag.java might break someone's code and 5.3 is already in beta.

We can put a `// FUTURE remove "deamon"` comment, which we usually check before working on a new version, so that we can remove it in the future.

sbleon:
Sounds good, Igal. I thought that daemon threads were a brand new feature in 5.3, but now I see that they've been around for 3 years! I agree that breaking changes should be avoided.

Do you know if there's a deprecation warning mechanism in Lucee? That would be a nice thing to have, but I couldn't find anything with a quick search.

isapir:
@sbleon Unfortunately there is no formal process in place for deprecation.  We have discussed it many times but nothing concrete came of that.  We should definitely set one up though, I agree.

bdw429s:
Usually when you ask the community about deprecating anything, the only response you get is "Wait, you can't _remove_ anything!!" :)

isapir:
Right, which is nonsensical if you ask me.

Major versions may have breaking changes IMO, though Removal should be preceded by Deprecation first, so at least 2 major versions to remove.

Python is the best example for a language that came back from the dead, and Python 3 is completely incompatible with Python 2.  If they can do that, then I don't see why we can't.

It will make the code cleaner and in some cases even faster, e.g. why test for a misspelled word like `deamon` in future versions?

isapir:
Having said that _here_... If you want to continue this discussion then let's, at the risk of starting WW3, do it in dev.lucee.org ;)

PR #538, opened 2018-07-13
FIx for LDEV-1912

cfmitrah:


micstriit:
manually adapted that change to branch 5.3 (not 5.2)
https://github.com/lucee/Lucee/commit/eeb20d6bae205bfbf1ee8d2009b06b20685235e1

PR #549, opened 2018-08-17
Fix for LDEV-1954

cfmitrah:


zspitzer:
this addresses a different problem, don't display the modern debug when all three tabs are disabled, as mentioned in https://luceeserver.atlassian.net/browse/LDEV-1865

PR #563, opened 2018-08-28
Changes in LDEV-1954

cfmitrah:


zspitzer (changes requested):
Typo: Mordern -> Modern, and shouldn't that be stText localised?

I still think this is unnecessarily complex, if no debugging template is configured, this view should just default to the Modern template

cfmitrah:
@zspitzer,

OK, I will configure to modern template if no debugging is configured.

PR #571, opened 2018-09-06
Fix for LDEV-1987

cfmitrah:


cfmitrah:
Added the fix in https://github.com/lucee/Lucee/pull/590

PR #585, opened 2018-10-04
Text Change: can only BE invoked (LDEV-2020)

mkasberg:
There is a small typo in the exception message. It used to read:

> Local scope can only invoked inside a function

It should read:

> Local scope can only BE invoked inside a function

This fixes [LDEV-2020](https://luceeserver.atlassian.net/browse/LDEV-2020).


cfmitrah:
Hi @mkasberg,

Thanks for PR. Merge it ASAP

PR #590, opened 2018-10-12
Fix for LDEV-2036

cfmitrah:


cfmitrah:
Added the tickets related to this fix
 LDEV-1965 , 
LDEV-1987, 
LDEV-2036 
LDEV-2037

PR #591, opened 2018-10-15
Fixes LDEV-784

Yamaha32088:


michaeloffner:
without digging to deep yet, i see 3 problems that need to be addressed
1. In KeyConstants we have this 
"public static final Key _csrf_token = KeyImpl._const("__csrf_token");"
the name ALWAYS must be identical with the value, expect that the name has a "_" prefix.
This is because the compiler reads this files and hardcodes keys found in the bytecode directly.
With this in KeyConstant, CFML code like this
<cfset x=session.csrf_token>
would be translated to
<cfset x=session.__csrf_token>
I'm not sure what this is about, so i don't know what the right key name is, but in any case the following rule apply
value=="_"&name

2. this code is also a problem
if(csrfTokens != null) {
	this.tokens = (Map<String, String>) csrfTokens.getValue();
it assumes that "csrfTokens" is a Map, what not necessary is the case so we need at least an "instanceof".

3. Also we never store Maps in structures visible to the user for one simple reason, Java maps are case sensitive (when using string keys), Lucee structs are not.
so the token should use a proxy class like this
MapAsStruct.toStruct(map,false)

Yamaha32088:
@michaeloffner thanks for the tips. When I get some time next week I will look into this again since most of what I did here has escaped my memory.

Yamaha32088:
@MarkSummers Looks like it wants you to sign the CLA, not sure if you can see that on your end or not. @michaeloffner this pull request has been updated with the changes you requested. Can you please take a look at it again? Thanks

PR #592, opened 2018-10-16
Added AsynchronousEvents.cfc in core

cfmitrah:


core/src/main/cfml/context/gateway/AsynchronousEvents.cfc
```diff
@@ -0,0 +1,90 @@
...
+
+    public function getState() {
+
+        return variables.state;
+    }
+
+
+    public function sendMessage( struct data={} ) {
+
+        return "sendGatewayMessage() has not been implemented for the event gateway [TaskGateway]. If you want to modify it, please edit the following CFC:"& expandpath("./") & "TaskGateway.cfc";
```
natewaite: Asynchronous Gateway requires the use of sendGatewayMessage my example has an example of how it could possibly be coded https://github.com/natewaite/Lucee/commit/03f169b7fed59d9e96c5fb4140343319f01aad35

natewaite (changes requested):

core/src/main/cfml/context/gateway/AsynchronousEvents.cfc
cfmitrah: @natewaite, 

Thanks for the info 

Changed sendmessage function based on your commit

michaeloffner:
@cfmitrah if still valid can you please redirect this to 6.0

michaeloffner:
@cfmitrah if that still makes sense, please make a new PR to Lucee 6.1

PR #595, opened 2018-10-23
Diff member function for dateDiff incompatible with ACF LDEV-2044

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2044


michaeloffner:
will not be added before Lucee 6, see cxomment in ticket

PR #614, opened 2018-11-29
Fix for LDEV-2014

goochjj:
building on @cfmitrah  fixes

### DIFF
--- a/core/src/main/java/lucee/commons/cli/Command.java
+++ b/core/src/main/java/lucee/commons/cli/Command.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 
+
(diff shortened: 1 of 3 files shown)


michaeloffner:
In any case the solution should have only minor affect on existing code, `cfexecute` is highly platform dependent, any change takes the risk to break existing code. Because of that code that uses a string as argument should still be handled the same way as before (as close as possible). Mean we still "parse" string arrays the same way as before and also pass it into `Runtime.exec` the same way as before. Only in case the users passes an array, we use the other `exec` method.

michaeloffner:
i adapted ideas from you, but also mades some things differently. now everything is converted to an array using the existing toArray method to avoid conflicts with existing code, in addition i fixed an issue with join when no timeout is defined.

PR #617, opened 2018-12-11
Added a test case for LDEV-2104

cfmitrah:


michaeloffner:
testcase cannot use cfhttp to call itself, that does not work within jsr223

PR #622, opened 2019-01-02
LDEV-2129 - Changing to return input instead of NULL when empty.

andrew-dixon:
Fixing ACF compat issue by changing to return input instead of NULL when empty. Details on:

https://luceeserver.atlassian.net/browse/LDEV-2129


michaeloffner:
i could not accept the PR from Andrew Dixon directly, because that function is no longer in the core in Lucee 5.3, but i did adapt his change to the esapi extension that now holds this function
https://github.com/lucee/extension-esapi/commit/8f4d5fb0bf60ccae68adb8fc55603d79796ae79b

PR #626, opened 2019-01-31
Added a fix for LDEV-2142

cfmitrah:


michaeloffner:
i have choosen a different way to solve this by fixing the underlaying problem
https://github.com/lucee/Lucee/commit/899cff493eb0bbbce6b3cb1cd15373386902e84d

PR #627, opened 2019-02-05
Added a test case for LDEV-2148

cfmitrah:


michaeloffner:
this is not expected to work that way, please see comment in ticket

PR #636, opened 2019-02-20
Added a fix & testcase for LDEV-0946

cfmitrah:


michaeloffner:
@cfmitrah as you can see above the test case still fails for this...

cfmitrah:
@micstriit  Due to conflict I close this PR and came with a new PR(https://github.com/lucee/Lucee/pull/1118)

PR #639, opened 2019-03-01
Added a fix for LDEV-2170

cfmitrah:


michaeloffner:
manually adapted the change, because an auto merge was not possible
https://github.com/lucee/Lucee/commit/8225d69485d812f48a0e2587ede60120c644bc2c

PR #647, opened 2019-04-02
Added a testcase for LDEV-1009

cfmitrah:


cfmitrah:
Wrongly committed with some other file so I closed this pull.

PR #649, opened 2019-04-02
Fixes LDEV-2136

bdw429s:
This change does the following

*  Follows redirects when downloading extension providers
* Doesn't error if an extension provider doesn't have the package since it needs to keep looping over all providers until it finds the extension
* Better logging for debugging errors
* Better status code checks for providers that return stuff like 203 instead of 200


bdw429s:
Thx!

PR #661, opened 2019-04-10
error stacktrace generation performance LDEV-2230 

DominicWatson:
Do not lookup classes in all mappings just to get full paths for stack trace. It both doesn't work and is very expensive. Use methods found elsewhere to get full path.

https://luceeserver.atlassian.net/browse/LDEV-2230


zspitzer:
@DominicWatson can you file this again against the 5.3 branch?

michaeloffner:
i have adpeted the change manually https://github.com/lucee/Lucee/commit/6612b5dcd74e7eefdddce9228486bd3decc6d3b7

@DominicWatson thanks for your work on this, sorry it took that long to "merge"

PR #668, opened 2019-04-16
Added a fix

cfmitrah:


michaeloffner:
the reason we use [ ] and not " or ' is to reduce conflicts in code that are using the error messages in javascript,json or other places where " and ' have a special meaning. so adding ' is the wrong way to go

PR #669, opened 2019-04-17
cfquery returnType Array return NULL whereas returnType Query return empty string with Partially NULL Support LDEV-2236

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2236


zspitzer:
fixed by https://github.com/lucee/Lucee/commit/dc931b72584042005a7d66a7feb0908a7290e84d

michaeloffner:
duplicate

PR #670, opened 2019-04-17
cacheGetAllIds argument order differs from ACF LDEV-2231

cfmitrah:
https://luceeserver.atlassian.net/browse/


michaeloffner:
this PR does not solve the issue, even the argument is named "cacheName" it is not, it is a region definition. we need to add region support in that context to solve the ticket.

PR #675, opened 2019-05-02
improve debugging execution time readability LDEV-2262

zspitzer:
- to make the report easier to scan don't show any 0 or 0.0 values, via a customNumberFormat which doesn't display 0.0 values
- added nowrap to the filter column for a more compact display (try to only have single lines per row)
- don't repeatedly show the "Stacktrace:" prefix, it's very common, only show the less common "Method:"
- removed unnecessary font tags (the TD already has the style)

https://luceeserver.atlassian.net/browse/LDEV-2262


isapir (changes requested):
The problem here is that the decimal delimiter might not be a ".", e.g. in a German locale it is a ",".  Using LSNumberFormat() might be better.

zspitzer:
this was already addressed via https://github.com/lucee/Lucee/pull/676 I guess the same review comments may apply there

PR #678, opened 2019-05-14
Replace by struct multiplies when value contains key LDEV-2273

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2273


zspitzer:
fixed by https://github.com/lucee/Lucee/commit/c9e3dc26273c48f479dc1e88c3206af706ed1449

PR #679, opened 2019-05-15
Added a fix for LDEV-2282

cfmitrah:


michaeloffner:
CFX still is supported! only C++CFX support is removed!

PR #682, opened 2019-05-16
improve error messages 

zspitzer:


isapir:
Not sure why the conflicts since the changes are limited in scope.  We will have to merge this manually or resolve in a git client (as opposed to online on github.com).

If you can generate a new PR it will save time but not necessary.

zspitzer:
https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/lucee/runtime/tag/Zip.java is 404, as it's been moved to an extension

I'll just remove it from the PR. I already submitted the same changes as PR for the extension https://github.com/lucee/extension-compress/pull/2 

Directory is conflicting, because a different PR from me was already accepted addressing the same items

PR #691, opened 2019-06-11
Cfschedule list returns wrong urls LDEV-2309

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2309


michaeloffner:
we have to be carefull that this value is not used alsewhere and in that case the port is added. in worst case scenario the port is added twice

core/src/main/java/lucee/runtime/tag/Schedule.java
```diff
@@ -579,7 +579,7 @@ private String printUrl(URL url) {
 		if (qs == null) qs = "";
 		else if (qs.length() > 0) qs = "?" + qs;
 
-		String str = url.getProtocol() + "://" + url.getHost() + url.getPath() + qs;
+		String str = url.getProtocol() + "://" + url.getHost() + ":" + Caster.toString(HTTPUtil.getPort(url)) + url.getPath() + qs;
```
zspitzer: @cfmitrah let's only add the port if non standard, i.e. not 443 with https or 80 with http?

zspitzer (changes requested):

core/src/main/java/lucee/runtime/tag/Schedule.java
cfmitrah: @zspitzer I've modified the fix to append the port if it is non-standard and also added the testcase

PR #694, opened 2019-06-13
cfdump does not expand when loaded via ajax LDEV-2210

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2210


michaeloffner:
ticket is already set to resolved  and this ticket has a conflict, so i assume this is a duplicate, if that is not the case, @cfmitrah please open a new ticket and a new PR for this 
https://luceeserver.atlassian.net/browse/LDEV-2210

PR #711, opened 2019-07-03
Added a fix for LDEV-2352

cfmitrah:


michaeloffner:
sorry, but you cannot do it that way,see here
https://github.com/lucee/Lucee/commit/d2257949ac618444e490b6fd95787065522b3e8b

PR #713, opened 2019-07-05
Can't verify built-in extension providers  LDEV-2327

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2327


michaeloffner:

PR #714, opened 2019-07-05
cfqueryparam list="true" inconsistency LDEV-2343

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2343


michaeloffner:
you cannot simply delete the type

michaeloffner:
@cfmitrah is this PR still valid, the only thing it does is NOT clone the type when a cloning is happening. to be honest this makes no sense to me and not cloning the type seems wrong. feel free to reopen the ticket with an explanation why not cling the type makes sense and solves the issue.

PR #715, opened 2019-07-05
Added a fix for LDEV-2344

cfmitrah:


michaeloffner:
sorry, but this simply breaks the functionality what will cause a lot of problems elsewhere, for example QoQ will no longer work.

PR #717, opened 2019-07-06
debugging improvements and bug fixes LDEV-2364

zspitzer:
- classic and comment templates no longer worked (just defaulted to modern)
- no access control, anyone can read a debug log if they know the log id, does cookie matching between the request and the log. Switching to a guid fixes this. See patch #723
- no guarantee that the correct debugging log was being displayed (fell back to last log if not found)
- updated the comment and classic templates to implement readDebug method like the modern template
- improved exception information for debug.cfc for unimplemented methods (getId, getLabel, getDescription) which the admin configure template interface relies on
- show problem template path in exception message for unimplemented methods in Debug.cfc
- show the debug execution time for the modern template (it's slow)
- duplicate the debug log pages data, as the modern template modifies the query and subsequent views of the logs via admin included the changes from the last view, resulting in very wrong statistics

https://luceeserver.atlassian.net/browse/LDEV-2364

### DIFF
--- a/core/src/main/java/resource/context/admin/debug/Classic.cfc
+++ b/core/src/main/java/resource/context/admin/debug/Classic.cfc
@@ -48,6 +48,9 @@ string function getid(){
 	return "lucee-classic"; 
 }
 
+string function readDebug(struct custom, struct debugging, string context){
+	output(argumentcollection=arguments);
+}
 
 void function onBeforeUpdate(struct custom){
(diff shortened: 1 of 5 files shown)


zspitzer:
huge Performance increase with the last commit, running the debugging on debugging is useful to see implicit variables access problems. one big init request went from 20s to 1.5s with this commit

PR #719, opened 2019-07-10
Added a fix for LDEV-2234

cfmitrah:


zspitzer:
this error message would be clearer if it mentioned it relates to sending a mail

cfmitrah:
There is some error in this PR so I close this ticket. Come with a new PR.

PR #723, opened 2019-07-17
debug log ids shouldn't be predictable  LDEV-2366

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2366


zspitzer:
fixed by another commit

PR #724, opened 2019-07-18
drastically improve load time for extension page LDEV-2377

zspitzer:


webonix:
What is the difference between var scoping a variable and using LOCAL  scope?

core/src/main/cfml/context/admin/ext.functions.cfm Line 175
ext not var scoped in else.

I have not seen functions with VARIABLES scope before; where does Lucee look for functions?

michaeloffner:
> What is the difference between var scoping a variable and using LOCAL scope?
> 
> core/src/main/cfml/context/admin/ext.functions.cfm Line 175
> ext not var scoped in else.
> 
> I have not seen functions with VARIABLES scope before; where does Lucee look for functions?

@webonix var and local. is a different for exactly the same, in ACF there is a difference, in Lucee not
functions are searched following the regular scope cascading.

PR #727, opened 2019-07-23
Added a fix for LDEV-2381

cfmitrah:


michaeloffner:
function works as expected, please see tickt for details.

PR #732, opened 2019-07-31
Added a fix for LDEV-2395

cfmitrah:


michaeloffner:
that change removes functionality, before it would still analyze the mime type based on the extension.

PR #735, opened 2019-08-09
Added a fix for LDEV-2306

cfmitrah:


michaeloffner:
@isapir do we not miss a line break after comment?

michaeloffner:
we can solve it that way, the dump already has the path, it simply misses the line number, in addition we need to solve this within the QueryImpl object

PR #737, opened 2019-08-19
Added a fix for LDEV-1997.

cfmitrah:


coffeeburrito (changes requested):
The port should only be hidden if 80 or 443. The getRequestUrl() function still needs to append the port in other circumstances (e.g. 8080, 8443, etc.)

cfmitrah:
I've close this PR and create new PR (https://github.com/lucee/Lucee/pull/1046)

PR #744, opened 2019-08-25
grammar improvements

zspitzer:
a error -> an error  (and so on, lots of missing an's)
throwed - > thrown
constellation - > combination

mostly in java doc, some in error messages and tag / functions definitions


zspitzer:
going to break this up into several PRs

PR #757, opened 2019-09-16
LDEV-2332 add a feature switch to turn off thread.stop()

DominicWatson:
Allow systems to set a `-Dlucee.conf.disable.thread.stop=true` environment variable to stop Lucee from attempting to brute force stop threads. Effectively turning off request timeouts.

See https://luceeserver.atlassian.net/browse/LDEV-2332

### DIFF
--- a/core/src/main/java/lucee/commons/io/SystemUtil.java
+++ b/core/src/main/java/lucee/commons/io/SystemUtil.java
@@ -4,17 +4,17 @@
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either 
+ * License as published by the Free Software Foundation; either
  * version 2.1 of the License, or (at your option) any later version.
- * 
+ *
  * This library is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  * Lesser General Public License for more details.
- * 
- * You should have received a copy of the GNU Lesser General Public 
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


bdw429s:
@DominicWatson Would it be better to make this part of the request timeout setting?  Such as a value of 0 or -1?

DominicWatson:
@bdw429s yes, and no. The issue with requestTimeout is that a developer can override it at any point in the request. You want a way for system administrators to disable the feature altogether.

Also, this prevents _any_ code from using Lucee's thread stopping code (which absolutely should be avoided).

Yes - there should be something better that allows you to specifically **disable** timeouts, and not just set them very high (setting them to zero in Lucee just has the effect of setting it really high). Ideally this should be configuration option (Admin + env vars, etc.).

But, right now the priority I think should be to prevent `thread.stop()` from happening at all. This is really a patch for the problem so that we can get stability back to our servers.

michaeloffner:
please instead of this
System.getProperty("lucee.conf.allow.thread.stop", "true") == "true";
do this
Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.conf.allow.thread.stop", null), false);

isapir:
I also don't like the name space of the property: `lucee.conf.allow.thread.stop`.

I think that it should be along the lines of `lucee.threads.stoppable` or `lucee.enable.thread.stop` or something else that is more consistent with the other properties that we already have.

DominicWatson:
Cool, yes - I did a quick search any didn't find anything so went with that (I saw the `lucee.conf...` namespace somewhere in the docs).

I think perhaps, `lucee.enable.request.timeout.monitor` + some more succinct logic that completely disables the Controler from watching for request timeouts might be preferable. I think that is less obscure a concept than "stopping thread.stop() in Lucee".

DominicWatson:
Going to close this and submit a new one when ready. This clearly isn't quite right.

PR #758, opened 2019-09-19
show lucee loader version on overview page LDEV-2459

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2459


zspitzer:
see also https://github.com/lucee/Lucee/pull/1173

PR #761, opened 2019-09-24
Added a fix for LDEV-2382

cfmitrah:


cfmitrah:
The LDEV-2382 issue was fixed in ticket LDEV-3042 and also due to some conflicts I close this PR

PR #771, opened 2019-10-11
Added a fix for LDEV-2517

cfmitrah:


michaeloffner:
sorry does not work for installed extensions, in that case fallback to "local".

PR #772, opened 2019-10-12
Updated README.md file with badges to show activity in community

sbruinsma:


DominicWatson:
@michaeloffner 

May I suggest that PRs to **master** are a good standard to follow as it keeps it simple for contributors. It is also logically correct, unless the contributor knows they are working on something that is still in development.

To accept a PR, you then personally merge it into your upcoming release branch and mark the PR as 'Merged into upcoming release' (or whatever labels in Github you choose to use) with a nice green colour to show everybody that it has been accepted. 

Finally, when the release becomes final, the PR logically closes as it has finally made it to stable Lucee.

This is what we do in Preside (similar volume of PRs) and it works really well. Benefits:

* Easy process for contributors to follow (no need to guess what branch to merge into)
* Contributors get their work attributed
* No more 'Rejected' PRs for work that was actually accepted

Here's what Preside PRs look like currently (I'm about to do a release, and all the open PRs will then close):

image

PR #776, opened 2019-10-18
Added a fix for LDEV-2518

cfmitrah:


zspitzer:
Can you split the test case for LDEV-2524 out into a separate PR?

The error messages shouldn't be hardwired, they should be localized, i.e added to
https://github.com/lucee/Lucee/blob/5.3/core/src/main/cfml/context/admin/resources/language/en.xml

As you have removed the individual cfoutput tags, can you re-indent the code?

PR #788, opened 2019-11-11
DeserializeJSON - numbers with lots of trailing zeroes casted to string LDEV-1973

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1973


michaeloffner:
we should simply use BigDecimal instead of this

PR #791, opened 2019-11-14
Added a fix for LDEV-2487

cfmitrah:


michaeloffner:
sorry but i have to reject this, that will do more harm than good. the comparator is locale based and that is the issue here, for more details see ticket.

PR #795, opened 2019-11-16
Add Support for elastic apm agent - LDEV-2572

xdecock:
This simply allows https://github.com/elastic/apm-agent-java to work correctly,

see bug : 

https://github.com/elastic/apm-agent-java/issues/885

We're currently working on an agent plugin


xdecock:
This is not needed anymore, latest version seems to manage injection without being an osgi loader

PR #796, opened 2019-11-16
use interrupt instead of stop to avoid thread leak - LDEV-2573

xdecock:
This uses the interrupt thread method that will allow the thread to leave the io blocking task instead of the deprecated stop();

this seems to avoid the thread leak.


michaeloffner:
Sorry we cannot adapt that PR, it would be fantastic if we could simply replace "stop" with "interrupt", bad sadly the "interupt" is very limited and not a replacement. but we do in most places first an interupt and only if that not works out, we go on with the stop...
But thanks a lot for your input!

PR #797, opened 2019-11-16
LDEV-2515 Prevent unwanted OSGi Downloads

DominicWatson:
This PR does three things:

1. Fixes an issue that leads to OSGi bundles needlessly being downloaded when they already exist in the Lucee core jar (this is the change around `.renameTo()` which silently fails in some situations)
2. Adds a flag to Lucee to allow administrators to stop Lucee from downloading Bundle files at all, `-Dlucee.enable.bundle.download=false`.
3. Upgrades logs about downloading bundles from DEBUG to WARN (and tweaks the language a little)

It changes the Loader. Yes. Unfortunately there is nothing we can do about that, the bug is in the Loader logic.

### DIFF
--- a/core/src/main/java/lucee/runtime/osgi/OSGiUtil.java
+++ b/core/src/main/java/lucee/runtime/osgi/OSGiUtil.java
@@ -3,17 +3,17 @@
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either 
+ * License as published by the Free Software Foundation; either
  * version 2.1 of the License, or (at your option) any later version.
- * 
+ *
  * This library is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  * Lesser General Public License for more details.
- * 
- * You should have received a copy of the GNU Lesser General Public 
+ *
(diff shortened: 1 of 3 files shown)


bdw429s:
Loader changes mean nothing to users on CommandBox or CommandBox-powered docker images as there is no such thing as an in-place upgrade.  Every server start is a full fresh war.  Please don't let people on legacy installations of Lucee hold up loader changes.  I say, put the fixes in now and people will get them as soon as they update their core jars.

zspitzer:
Yeah, I'd say just fix the loader regardless of any schedule for loader changes. 

If people need the fix, it should be available

pfreitag:
Can we also add this setting as an environment variable? `LUCEE_ENABLE_BUNDLE_DOWNLOAD=false` easier to set in some cases (such as docker)

michaeloffner:
@DominicWatson @zspitzer  loader changes as this are fine, as long they do not affect the core in any way.
what you never can do in the loader is changing any interface the core then uses. that would break any installation using an older loader. also any loader update needs a new loader version set.
I will accept the PR and make the necessary adjustments including the variable handling.

michaeloffner:
@DominicWatson thanks for your work on this
https://luceeserver.atlassian.net/browse/LDEV-2475

DominicWatson:
> @DominicWatson @zspitzer loader changes as this are fine, as long they do not affect the core in any way.
> what you never can do in the loader is changing any interface the core then uses. that would break any installation using an older loader.

Aha, great to understand that a bit better, thank you for explaining. And for merging :)

PR #799, opened 2019-11-17
Ldev 2332 Alternative to #798 building on top of #759

xdecock:
Basically,

Adapt the bytecode change in #798  to build on top of @DominicWatson proposal. #759 

The iteration count before checking the timeout is set arbitrarely to 10K, this will need some test and tuning.

This works by adding a call to PageContextUtil.checkRequestTimeout at the end of loop opcode


xdecock:
Something a bit better we could do is probably to allow preempt = false on the loop tag to avoid the performance hit if we are sure we do not need the protection.

Could even be set as default on the application level, but imho this can be done in a future iteration, i'm not sure a lot of people will benefit from the performance vs security tradeoff

xdecock:
Yet another note after having that code running.

We probably need to do the same for the "retry operation", as it might trigger a loop

and after the end of the loop

PR #800, opened 2019-11-19
LDEV-2578 reduce info logging

DominicWatson:
While watch _all_ logs in a docker container, we noticed a lot of logs that we think might be better off at DEBUG level. Also a couple of logs that could be on a single line and worded more informatively (particularly useful when aggregating multiple logs into one system).


michaeloffner:
https://github.com/lucee/Lucee/commit/d21c39082a576aa67b765d87b6315c0d57c8eb63

PR #803, opened 2019-11-22
Added a test case for LDEV-2586

cfmitrah:


michaeloffner:
testcase PR should not fix code ...

michaeloffner:
i disagree on this fix, there should be a difference between D and d, it is not the same in that case. upper D should be untouched.

michaeloffner:
i will accept PR but revert the change in LSParseDateTime

PR #806, opened 2019-11-28
Port Performance improvements from Lucee 5.3 to Lucee 5.2

migueltarga:


zspitzer:
Lucee 5.2 is no longer maintained

PR #812, opened 2019-12-05
fix cfqueryparam null=true not ignoring sqltype LDEV-2581

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2581


michaeloffner:
@cfmitrah this seem to be a duplicate, code looks completely different today and the ticket is resolved. please check it there is still a need for this improvment, if so create a new ticket with a new PR.

PR #813, opened 2019-12-05
queryAddRow with a struct of column keys, handles arrays wrong LDEV-1206

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1206


zspitzer:
test case is was already merged in https://github.com/lucee/Lucee/pull/182

PR #824, opened 2020-01-08
Added a fix and test case for LDEV-2632

cfmitrah:


michaeloffner:
we cannot simply return null in that case if null support is not enabled.

michaeloffner:
that code change completely changes how arrays handles null values, if not full null support a null value is equal to not existing

PR #825, opened 2020-01-08
fix admin button display LDEV-2633

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2633


zspitzer:
see also https://github.com/lucee/Lucee/pull/990

PR #826, opened 2020-01-14
Documentation for cfquery updated

richardherbert:
Attribute cacheid (not implemented) added


michaeloffner:
adapted PR manually, because it was targeting the wrong branch "master" instead of "5.3".
Also the status need to be "unimplemented" not "not implemented"
https://github.com/lucee/Lucee/commit/2d68d89b87d2c0e1eea7fb2b083fdf4ba3e1e264@

@richardherbert thanks a lot

PR #830, opened 2020-01-22
Added a test case for postgreSQL

cfmitrah:


michaeloffner:
@cfmitrah i added the PR even we already have a PostgreSQL test case at test/datasource/PostgreSQL.cfc

PR #835, opened 2020-01-30
Added a better error message for LDEV-2657

cfmitrah:


michaeloffner:
Sorry i have to reject this, this is the case for many functions, we need a more general approach for this, also your change would destroy the output when you use arraySlice not as a member function

PR #839, opened 2020-02-05
cfimap, add missing action to error message

zspitzer:
movemail wasn't included in the error message


zspitzer:
https://github.com/lucee/lucee-docs/pull/743

PR #848, opened 2020-02-14
update format of error messages to match convention #2

zspitzer:


core/src/main/java/lucee/runtime/type/UDFAddProperty.java
```diff
@@ -68,13 +68,13 @@ public Object _call(PageContext pageContext, Object[] args, boolean doIncludePat
 		// struct
 		if (this.arguments.length == 2) {
 			if (args.length < 2) throw new ExpressionException(
-					"The function " + getFunctionName() + " need 2 arguments, only " + args.length + " argment" + (args.length == 1 ? " is" : "s are") + " passed in.");
+					"The function [" + getFunctionName() + "] needs 2 arguments, only " + args.length + " argument" + (args.length == 1 ? " is" : "s are") + " passed in.");
```
isapir, line 71: was/were would be more consistent instead of is/are

PR #864, opened 2020-02-20
improve invalid access type error message

zspitzer:


isapir:
So we actually return in the error message the constants names instead of values?

> Component.ACCESS_PACKAGE, Component.ACCESS_PRIVATE, Component.ACCESS_PUBLIC, Component.ACCESS_REMOTE

I would think that it makes more sense to return the string representation of the values.

These values are integers, which is also not that great for the user, e.g. 
https://github.com/lucee/Lucee/blob/master/loader/src/main/java/lucee/runtime/Component.java#L45

I'm pulling this in as it was that way before, but perhaps it would make sense to improve on that.

zspitzer:
I only have seen the first error message, figured I'd update that odd one to match convention

PR #868, opened 2020-02-27
fix Error -with the 'Debugging - Settings' in WEB Admin  LDEV-2726

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2726


zspitzer:
what does secValue mean? docs are missing a description
https://docs.lucee.org/guides/Various/cfadmin.html

cfmitrah:
@zspitzer, With secValue, you can define the access rights for the settings that can be overridden in the "web administrator". If true means you can create mappings, data source, etc, false means "read-only" only view the settings used in server admin.

PR #905, opened 2020-03-27
sameFormFieldsAsArray does not convert fields to arrays if all inputs are empty LDEV-2795

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2795
https://luceeserver.atlassian.net/browse/LDEV-1619


core/src/main/java/lucee/runtime/type/scope/ScopeSupport.java
```diff
@@ -241,26 +241,20 @@ else if (curr instanceof String) {
 			if (isArrayDef) {
 				Array arr = new ArrayImpl();
 				arr.appendEL(curr);
-				arr.appendEL(value);
+				arr.appendEL(value);
```
jbampton:
```suggestion
				arr.appendEL(value);
```

cfmitrah:
@zspitzer update testcase from https://github.com/lucee/Lucee/pull/1352

PR #914, opened 2020-04-03
persist debug=true in session for admin

zspitzer:
let's you turn on debugging and leave it on

### DIFF
--- a/core/src/main/cfml/context/admin/admin_layout.cfm
+++ b/core/src/main/cfml/context/admin/admin_layout.cfm
@@ -143,6 +143,8 @@
 </cfoutput>
 	<cfset thistag.generatedcontent="">
 </cfif>
-
-<cfparam name="url.debug" default="no">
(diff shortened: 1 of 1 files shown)


michaeloffner:
@zspitzer thanks!

zspitzer:
@michaeloffner oops, this conflicts with the persistent debug template selection, which also uses `session.debug`

```
5.3.7.14-SNAPSHOT Error (expression)
--
Message | can't assign value to an Object of this type [String] with key template 
The Error Occurred in/admin/debugging.logs.detail.cfm: line 2
called from /admin/debugging.logs.cfm: line 151
called from /admin/web.cfm: line 467
```

https://github.com/lucee/Lucee/blob/5.3/core/src/main/cfml/context/admin/debugging.logs.detail.cfm#L2

zspitzer:
Improve that error message https://github.com/lucee/Lucee/pull/934

PR #917, opened 2020-04-06
Extensions with no "release" update available are showing as requiring an update LDEV-2716

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2716


andrew-dixon:
@michaeloffner - If this pull request is ok, can it be merged, the admin keeps telling me there are extension updates when there are not, it is very annoying 😄

PR #921, opened 2020-04-08
Added a fix for LDEV-2834

cfmitrah:


michaeloffner:
this change no longer works as expected, because the upadte url got changed to https

PR #922, opened 2020-04-11
Added from email address to UI for sending test emails

penhorwood:
On the web.cfm -> email - mail server connections,
added code to use a user defined from email address in the send test email user interface.
Then added code to the send test email to use this new fromMail value.


michaeloffner:
@penhorwood thanks

PR #923, opened 2020-04-12
admin, show settings, export Application.cfc code by default

zspitzer:
there's no need for it to be hidden by default

Before
image

After
image


michaeloffner:
@zspitzer thanks!

PR #931, opened 2020-04-23
Fix LDEV-2778

kabutotx:
Fix rounding error if number parameter was done as mathematical statement.


michaeloffner:
it's a good idea, but i'm really not sure if we should go down this route. Math in every computer language is limited to the binary representation of a number in memory.

kabutotx:
Left some comments on the ticket.
I noticed BigDecimal.ROUND_HALF_UP is deprecated in java 11 and we should be using RoundingMode.HALF_UP from java.math.RoundingMode

kabutotx:
Closing since round() is giving same answer (3.07) as numberFormat() and decimalFormat() on ACF.
Other Lucee functions are incompatible with ACF. See LDEV-2863.

PR #934, opened 2020-04-24
improve error message for assigning struct

zspitzer:
map and object aren't really cfml terms either.... struct would be more better?


michaeloffner:
thanks for your work on this!

PR #935, opened 2020-04-28
Ldev 1236

mircea-botez:
- added samesite attribute to Cookie


zspitzer:
Nice!

There's a few cookies in the Lucee admin which could do with some samesite lockdown 
https://github.com/lucee/Lucee/blob/5.3/core/src/main/cfml/context/admin/web.cfm#L83

but better via the Application.cfc, 

`
this.tag.cookie.samesite="strict";
`

PR #937, opened 2020-04-28
LDEV-2568-fixed int issue for expires attribute in cookies

mircea-botez:


michaeloffner:
adapted the change manually

PR #939, opened 2020-04-29
Rounding in functions is different than Coldfusion LDEV-2863

kabutotx:
Makes rounding compatible to ACF.

round: fixes error and change to RoundingMode as BigDecimal round is being depreceated.
numberFormat: remove influence and change to BigDecimal
decimalFormat (Caster): change to BigDecimal to fix errors in Math.round

Optional
LSCurrencyFormat: removes influence to make compatible with ACF but gives wrong rounding like ACF

https://luceeserver.atlassian.net/browse/LDEV-2863


michaeloffner:
this PR is obsolete because we switched to BigDecimal with Lucee 6, Lucee 6 is no much more percise with math operations and for sure differes to ACF. compatibility to ACF would be a big downside.

PR #940, opened 2020-04-29
improve UDF error messages

zspitzer:


michaeloffner:
thanks for your work on this!

PR #942, opened 2020-04-30
use async=false when sending a test email LDEV-2873

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-2873


michaeloffner:
thanks

PR #952, opened 2020-05-15
Add iso format for DateTimeFormat LDEV-2881

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2881


zspitzer:
already fixed https://github.com/lucee/Lucee/blob/6.0/core/src/main/java/lucee/runtime/functions/displayFormatting/DateTimeFormat.java#L94

PR #955, opened 2020-05-25
enable samesite=strict for admin

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-1236
https://luceeserver.atlassian.net/browse/LDEV-2900


zspitzer:
bit of an edge case
`this.tag.cookie.sameSite = "strict";`
isn't used with 
`<cfparam name="cookie.lucee_admin_lang" default="en">`

PR #958, opened 2020-06-08
struct.delete() should return the struct LDEV-2915

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2915

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/struct/StructDelete.java
+++ b/core/src/main/java/lucee/runtime/functions/struct/StructDelete.java
@@ -39,7 +39,7 @@ public final class StructDelete extends BIF {
 	}
 
 	public static boolean call(PageContext pc, Struct struct, String key, boolean indicatenotexisting) throws TemplateException {
-		if(indicatenotexisting && !struct.containsKey(key)) throw new TemplateException("Cannot delete item with key " + key, "The key doesn't exists.");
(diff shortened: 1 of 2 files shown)


core/src/main/java/lucee/runtime/functions/struct/StructDelete.java
```diff
@@ -33,11 +34,12 @@
...
 	private static final long serialVersionUID = 6670961245029356618L;
 
-	public static boolean call(PageContext pc, Struct struct, String key) {
+	public static boolean call(PageContext pc, Struct struct, String key) throws TemplateException {
 		return call(pc, struct, key, false);
 	}
 
-	public static boolean call(PageContext pc, Struct struct, String key, boolean indicatenotexisting) {
+	public static boolean call(PageContext pc, Struct struct, String key, boolean indicatenotexisting) throws TemplateException {
+		if(indicatenotexisting && !struct.containsKey(key)) throw new TemplateException("Cannot delete item with key " + key, "The key doesn't exists.");
```
jbampton:
```suggestion
		if(indicatenotexisting && !struct.containsKey(key)) throw new TemplateException("Cannot delete item with key " + key, "The key doesn't exist.");
```

jbampton (changes requested):

PR #978, opened 2020-07-06
Added a fix for LDEV-27

cfmitrah:


michaeloffner:
i disagree on this fix, please see my first comment in the ticket, general parsing only should be used when local is "US", if we change that, this could break a lot a existing code. Do you disagree ?

cfmitrah:
@micstriit , We should consider the existing code too. So, OK for your intention of this issue. Is there any way to prevent the issue if any user faces this error?

michaeloffner:
to be honest i would prefer to have that setting always on "false", ls* functions should not support this stupid unlogical collections of supported parsing masks, this should be limited to non "ls" date functions.
But for compatibility we do support it anyway.
Why you see the need for this?

cfmitrah:
@michaeloffner, Just for my clarification only about the issue.

michaeloffner:
@cfmitrah the underlying ticket for this is set to resolved. if that is still an issue that need to be addressed, please open a new ticket (that links LDEV-27) and create a new PR for Lucee 6.1

PR #986, opened 2020-07-22
fix updating a datasource LDEV-2978

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2978


zspitzer:
this caused a regression https://github.com/lucee/Lucee/pull/1303

PR #988, opened 2020-07-27
https://luceeserver.atlassian.net/browse/LDEV-2883 - made arguments o…

mircea-botez:
…ptional and added a test case


michaeloffner:
added the changes manually, because of conflicts
for detail see ticket

PR #990, opened 2020-07-31
Added some enhancement fixes in admin application search box

cfmitrah:


michaeloffner:
text should coming from <language>.xml

cfmitrah:
@michaeloffner I moved the message to the XML file

PR #996, opened 2020-08-04
Added a samesite column in http results query

cfmitrah:


michaeloffner:
i nowhere see that a value is set for that column, what do i miss?

zspitzer:
doesn't parse cookie populate the column?

https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/lucee/runtime/tag/Http.java#L1452

michaeloffner:
i assume it is set here?
https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/lucee/runtime/tag/Http.java#L1474

PR #1000, opened 2020-08-12
Added a fix for LDEV-3001

cfmitrah:


zspitzer:
I improved the error messages in the same file https://github.com/lucee/Lucee/pull/1001

PR #1002, opened 2020-08-13
Fix for LDEV-3008: Upload new extension UI Asks for Wrong File Type.

JayIsPainting:
Changed 3 language files to modify UI from asking for ZIP files when expecting LEX files.


zspitzer (changes requested):
File shouldn't be upper case?

JayIsPainting:
I tend to do that for input labels but I can certainly change it.

michaeloffner:
https://github.com/lucee/Lucee/commit/372cc048da2710a0067f649829107ceba19fbc8d

PR #1010, opened 2020-08-25
Added a block instead of black for LDEV-3017

cfmitrah:


michaeloffner:
https://github.com/lucee/Lucee/commit/6159b91234ea9dba01872a9c5473acd18cb24194

PR #1013, opened 2020-08-31
include bad email address in exception LDEV-2234

zspitzer:


michaeloffner:
@zspitzer i disagree on this, we try never to expose private info like this in error messages, you never want to expose any email address on the website and yes a lot of apps show error message to the user.

zspitzer:
I understand, is there a possibility to just log this out and not return it as a error message?

similar problem as https://luceeserver.atlassian.net/browse/LDEV-3012

It's currently very hard to go back and debug smtp errors as there is only the error but not enough contextual info

PR #1023, opened 2020-09-04
Added a fix for LDEV-3027

cfmitrah:


michaeloffner:
please see conversation in ticket

PR #1026, opened 2020-09-08
LDEV-3042 - Query of Query performance is very bad and single threaded for complex SQL.

bdw429s:
This pull is for ticket
https://luceeserver.atlassian.net/browse/LDEV-3042

It should also fix tickets
* https://luceeserver.atlassian.net/browse/LDEV-2541
* https://luceeserver.atlassian.net/browse/LDEV-1272

Improve the native Java implementation of QofQ to support
* group by clause
* having  clause
* aggregate functions
* Allow aggregates to reference nested operations including scalar functions
  * `ceiling( max( floor( yearsEmployed ) )+count(1) )`
* faster distinct support
* Fix buginess in unions were results aren't distinct by default (union distinct)
* Improve count() to support 
  * `count( all col )`
  * `count( 1 )`
  * `count( * )`
  * `count( 'literal' )`
  * `count( distinct col )`
  * `count( distinct scalarFunc( col ) )`
* Fix bugs in SQL parser that incorrectly requires only a single space between multi-word clauses
  * `is null`
  * `is not null`
  * `not in`
  * `not like`
  * `order by`
  * `group by`
* Remove single-threaded limitations
* Performance tune speed of queries

I have also added support for the following system props/env vars
* **lucee.qoq.hsqldb.disable**=true – Throw exception if native QoQ logic fails
* **lucee.qoq.hsqldb.debug**=true – Log message to WEB context's datasource log any time a QofQ "falls back" to HyperSQL. This could include just bad SQL syntax.

@michaeloffner @micstriit Please ask if you have any questions.  Every change here I have made for a reason-- many of them as part of my performance tuning to shave extra ms off of processing times in large loops.  Note there is a new TestBox spec that tests all the new functionality.

### DIFF
--- a/core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
+++ b/core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
@@ -70,6 +70,23 @@ public final class HSQLDBHandler {
 	Executer executer = new Executer();
 	QoQ qoq = new QoQ();
 	private static Object lock = new SerializableObject();
+	private static boolean hsqldbDisable;
+	private static boolean hsqldbDebug;
+
+	static {
+		try {
+			hsqldbDisable = Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.disable", "false"));
+		}
+		catch (PageException e) {
+			hsqldbDisable = false;
+		}
+		try {
+			hsqldbDebug = Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.debug", "false"));
+		}
+		catch (PageException e) {
+			hsqldbDebug = false;
+		}
+	}
 
 	/**
 	 * constructor of the class
@@ -248,7 +265,7 @@ public final class HSQLDBHandler {
 		try {
 			SelectParser parser = new SelectParser();
 			selects = parser.parse(sql.getSQLString());
-			QueryImpl q = qoq.execute(pc, sql, selects, maxrows);
+			QueryImpl q = (QueryImpl) qoq.execute(pc, sql, selects, maxrows);
 			q.setExecutionTime(stopwatch.time());
 			return q;
 		}
@@ -268,12 +285,12 @@ public final class HSQLDBHandler {
 		}
 
 		// Debugging option to completely disable HyperSQL for testing
-		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.disable", "false"))) {
+		if (qoqException != null && hsqldbDisable) {
 			throw Caster.toPageException(qoqException);
 		}
 
 		// Debugging option to to log all QoQ that fall back on hsqldb in the datasource log
-		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.debug", "false"))) {
+		if (qoqException != null && hsqldbDebug) {
 			pc.getConfig().getLog("datasource").error("QoQ [" + sql.getSQLString() + "] errored and is falling back to HyperSQL.", qoqException);
 		}
 
--- a/core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
+++ b/core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
@@ -32,6 +32,7 @@ import java.util.Iterator;
 import java.util.Set;
 
 import lucee.commons.db.DBUtil;
+import lucee.commons.io.SystemUtil;
 import lucee.commons.lang.ExceptionUtil;
 import lucee.commons.lang.SerializableObject;
 import lucee.commons.lang.StringUtil;
@@ -241,6 +242,8 @@ public final class HSQLDBHandler {
 		String prettySQL = null;
 		Selects selects = null;
 
+		Exception qoqException = null;
+
 		// First Chance
 		try {
 			SelectParser parser = new SelectParser();
@@ -250,7 +253,7 @@ public final class HSQLDBHandler {
 			return q;
 		}
 		catch (SQLParserException spe) {
-			// sp
+			qoqException = spe;
 			prettySQL = SQLPrettyfier.prettyfie(sql.getSQLString());
 			try {
 				QueryImpl query = executer.execute(pc, sql, prettySQL, maxrows);
@@ -260,8 +263,19 @@ public final class HSQLDBHandler {
 			catch (PageException ex) {}
 
 		}
-		catch (PageException e) {}
-		// if(true) throw new RuntimeException();
+		catch (PageException e) {
+			qoqException = e;
+		}
+
+		// Debugging option to completely disable HyperSQL for testing
+		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.disable", "false"))) {
+			throw Caster.toPageException(qoqException);
+		}
+
+		// Debugging option to to log all QoQ that fall back on hsqldb in the datasource log
+		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.debug", "false"))) {
+			pc.getConfig().getLog("datasource").error("QoQ [" + sql.getSQLString() + "] errored and is falling back to HyperSQL.", qoqException);
+		}
 
 		// SECOND Chance with hsqldb
 		try {
--- a/core/src/main/java/lucee/runtime/db/QoQ.java
+++ b/core/src/main/java/lucee/runtime/db/QoQ.java
@@ -31,6 +31,7 @@ import lucee.runtime.PageContext;
 import lucee.runtime.exp.DatabaseException;
 import lucee.runtime.exp.PageException;
 import lucee.runtime.op.Caster;
+import lucee.runtime.op.Decision;
 import lucee.runtime.op.Operator;
 import lucee.runtime.sql.QueryPartitions;
 import lucee.runtime.sql.Select;
@@ -244,8 +245,9 @@ public final class QoQ {
 			trgValues[cell] = expSelectsMap.get(headers[cell]);
 		}
 
-		// If have a group by, a distinct, or this is part of a "union" then we partition
-		if (select.getGroupbys().length > 0 || select.isDistinct()) {
+		// If have a group by, a distinct, or this is part of a "union", or has aggregates in the
+		// select list then we partition
+		if (select.getGroupbys().length > 0 || select.isDistinct() || (select.hasAggregateSelect() && select.getWhere() != null)) {
 			executeSinglePartitioned(pc, select, source, target, maxrows, sql, hasOrders, isUnion, trgColumns, trgValues, headers);
 		}
 		// This is a "normal" select with no partitioning
@@ -397,12 +399,12 @@ public final class QoQ {
 				for (int cell = 0; cell < headers.length; cell++) {
 					trgColumns[cell].set(target.getRecordcount(), getValue(pc, sql, source, row, headers[cell], trgValues[cell]));
 				}
+			}
 
-				// If this was a non-grouped select with only aggregates like select "count(1) from
-				// table" than bail after a single row
-				if (hasAggregateSelect) {
-					break;
-				}
+			// If this was a non-grouped select with only aggregates like select "count(1) from
+			// table" than bail after a single row
+			if (hasAggregateSelect) {
+				break;
 			}
 		}
 
@@ -559,6 +561,46 @@ public final class QoQ {
 		throw new DatabaseException("unsupported sql statement [" + exp + "]", null, sql, null);
 	}
 
+	/**
+	 * Accepts an expression which is the input to an aggregate operation.
+	 * 
+	 * @param pc
+	 * @param sql
+	 * @param source
+	 * @param exp
+	 * @param row
+	 * @return an array with as many items as rows in the source query containing the corresponding
+	 *         expression result for each matching row
+	 * @throws PageException
+	 */
+	private Object[] executeAggregateExp(PageContext pc, SQL sql, Query source, Expression exp) throws PageException {
+		Object[] result = new Object[source.getRecordcount()];
+
+		// For a literal value, just fill an array with that value
+		if (exp instanceof Value) {
+			Object value = ((Value) exp).getValue();
+			for (int i = 0; i < source.getRecordcount(); i++) {
+				result[i] = value;
+			}
+			return result;
+		}
+
+		// For a column, return the data in that column as an array
+		if (exp instanceof Column) {
+			return ((QueryColumnImpl) source.getColumn(((Column) exp).getColumn())).toArray();
+		}
+		// For an operation, we need to execute the operation once for each row and capture the
+		// results as our final array
+		if (exp instanceof Operation) {
+			for (int i = 0; i < source.getRecordcount(); i++) {
+				result[i] = executeOperation(pc, sql, source, (Operation) exp, i + 1);
+			}
+			return result;
+		}
+
+		throw new DatabaseException("unsupported sql statement [" + exp + "]", null, sql, null);
+	}
+
 	private Object executeOperation(PageContext pc, SQL sql, Query source, Operation operation, int row) throws PageException {
 
 		if (operation instanceof Operation2) {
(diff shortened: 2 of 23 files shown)


stillnet:
Looking forward to this fix. Just recently spent a day of troubleshooting & refactoring on a performance issue related to QoQ with aggregation.

bdw429s:
@stillnet I've done extensive performance tests on my new implementation and not only is it faster than the previous Lucee implementation in every test _especially_ under load, it's also faster than Adobe ColdFusion 2018 in most tests as well.  See my performance report here:
https://drive.google.com/file/d/12iqZhbIh62spdJBQC2MaOroiMf8D4l7L/view?usp=sharing

stillnet:
@bdw429s wow that is a huge improvement, nice work! :-)

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -73,205 +96,423 @@ public Query execute(PageContext pc, SQL sql, int maxrows) throws PageException
...
 			}
 		}
 
-		// Group By
-		if (select.getGroupbys().length > 0) throw new DatabaseException("group by are not supported at the moment", null, sql, null);
-		if (select.getHaving() != null) throw new DatabaseException("having is not supported at the moment", null, sql, null);
+		// Add first row of each group of partitioned data into final result
+		QueryImpl sourcePartition;
+		for (String partionKey: queryPartitions.getPartitions().keySet().toArray(new String[queryPartitions.getPartitions().size()])) {
+			sourcePartition = queryPartitions.getPartitions().get(partionKey);
```
michaeloffner: again 
`for(QueryImpl sourcePartition:...values())`

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -302,72 +543,132 @@ private Query getSingleTable(PageContext pc, Column table) throws PageException
...
+			// Simple return if we want all values
+			if (!returnDistinct && includeNull) {
+				return result;
+			}
+			// if we want to filter nulls or distinct the array
+			else {
+				Stream<Object> resultStream = Arrays.stream(result);
+				if (!includeNull) resultStream = resultStream.filter(s -> (s != null));
+				if (returnDistinct) resultStream = resultStream.distinct();
+				return resultStream.toArray(Object[]::new);
```
michaeloffner: that syntax will not work with Java 8 and we still compile with Java 8 for backward compatibility, so better symply do
`resultStream.toArray()`

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -302,72 +543,132 @@ private Query getSingleTable(PageContext pc, Column table) throws PageException
...
```
michaeloffner: see above

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -396,38 +697,51 @@ private Object executeOperation(PageContext pc, SQL sql, Query qr, Operation ope
...
+			else {
+				value = executeExp(pc, sql, source, operators[0], row);
+			}
 
 			// Functions
 			switch (op.charAt(0)) {
 			case 'a':
 				if (op.equals("abs")) return new Double(MathUtil.abs(Caster.toDoubleValue(value)));
 				if (op.equals("acos")) return new Double(Math.acos(Caster.toDoubleValue(value)));
+				https: // ortussolutions.slack.com/archives/GUEEFNMDX
```
michaeloffner: that url should break the compiler ...

core/src/main/java/lucee/runtime/sql/QueryPartitions.java
```diff
@@ -0,0 +1,311 @@
...
+		return partitions;
+	}
+
+	/**
+	 * Get array of grouped QueryImpl object
+	 * 
+	 * @return
+	 */
+	public QueryImpl[] getPartitionArray() {
+		return (QueryImpl[]) partitions.values().toArray();
```
michaeloffner: i see in a lot of places "QueryImpl" is used, even "Query" is sufficient, whenever possible go down the the interface that is sufficient. by using "Query" you make the code more flexible.

core/src/main/java/lucee/runtime/type/KeyImpl.java
```diff
@@ -192,7 +192,8 @@ public boolean equals(Object other) {
 			return key.equalsIgnoreCase((String) other);
 		}
 		if (other instanceof Key) {
-			return ucKey.equalsIgnoreCase(((Key) other).getUpperString());
+			// Both strings are guaranteed to be upper case
+			return ucKey.equals(((Key) other).getUpperString());
```
michaeloffner, line 196: as far as i remember I have choosen "equalsIgnoreCase" because it is faster, because "equalsIgnoreCase" argument is a string and the one from "equals" is an object. But we need to test this, maybe it also changed with never JVM.

core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
```diff
@@ -260,8 +263,19 @@ public QueryImpl execute(PageContext pc, final SQL sql, int maxrows, int fetchsi
...
 
 		}
-		catch (PageException e) {}
-		// if(true) throw new RuntimeException();
+		catch (PageException e) {
+			qoqException = e;
+		}
+
+		// Debugging option to completely disable HyperSQL for testing
+		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.disable", "false"))) {
```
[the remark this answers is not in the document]
bdw429s: @michaeloffner The reason I was checking it every time was because during my tests I would toggle the setting back and forth between tests without restarting the server.  I can make this a static variable though if we think users will not do that.

core/src/main/java/lucee/runtime/db/QoQ.java
bdw429s: @michaeloffner  I have no idea how that got there!  Must have been an accidental paste.  Removing...

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -30,33 +35,51 @@
...
 import lucee.runtime.type.QueryColumn;
+import lucee.runtime.type.QueryColumnImpl;
 import lucee.runtime.type.QueryImpl;
+import lucee.runtime.type.util.ArrayUtil;
 
 /**
  * 
  */
 public final class QoQ {
+	final static private Collection.Key paramKey = new KeyImpl("?");
```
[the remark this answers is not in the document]
bdw429s: @michaeloffner Sorry, I don't follow what your question is.  I created the Key here just to prevent the overheard of creating it in every place it was used.  We can put this in `KeyConstants.java`, but I wasn't sure what the criteria was for doing that.

core/src/main/java/lucee/runtime/sql/QueryPartitions.java
bdw429s: @michaeloffner Yes, I am aware that you always use the interface.  however, the issue I was running into is that not all the methods declared in `QueryImpl` are also declared in the `Query` interface.  `cutRowsTo()` is an example as well as the new overloaded `setAt()` method I added.  I understand changing the interface in a minor version is not an option so I did not touch the interface.  

Since you brought it up, I have changed all the types over to use the `Query` interface and I am casting to `QueryImpl` in the places it is required.  Ex:
```java
((QueryImpl) target).cutRowsTo(maxrows);
```
Is that an ok workaround?

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -73,205 +96,423 @@ public Query execute(PageContext pc, SQL sql, int maxrows) throws PageException
...
+		// ...and all of the new rows
+		for (int row = 1; row <= target.getRecordcount(); row++) {
+			queryPartitions.addRow(pc, target, row, true);
+		}
+
+		QueryImpl sourcePartition;
+		// Loop over the partitions and take one from each and add to our new target question for a
+		// distinct result
+		for (String partionKey: queryPartitions.getPartitions().keySet().toArray(new String[queryPartitions.getPartitions().size()])) {
+			sourcePartition = queryPartitions.getPartitions().get(partionKey);
```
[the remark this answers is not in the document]
bdw429s: Good point. I'm just a little unfamiliar with for loops over collections in Java.  I'm not sure why I didn't do that.  I will modify this.

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -73,205 +96,423 @@ public Query execute(PageContext pc, SQL sql, int maxrows) throws PageException
...
+				// ... add this row to our partitioned data
+				queryPartitions.addRow(pc, source, row, false);
+			}
+		}
+
+		// Now that all rows are partioned, eliminate partions we don't need via the having clause
+		if (select.getHaving() != null) {
+			// Loop over each partition
+			for (String partionKey: queryPartitions.getPartitions().keySet().toArray(new String[queryPartitions.getPartitions().size()])) {
+				// Eval the having clause on it
```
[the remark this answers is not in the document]
bdw429s: Good call.  I was concerned about removing keys from the Map as I looped over it, but this should be ok.  I will change it to use an `entrySet()`.  This is me being a little unfamiliar again with looping over Maps in Java.

core/src/main/java/lucee/runtime/db/QoQ.java
bdw429s: Changed.

core/src/main/java/lucee/runtime/type/KeyImpl.java
bdw429s: @michaeloffner No, a case-insensitive comparison should always be faster.  I found this bottleneck while profiling the code during my performance testing and it was internally changing the case of the strings.  Both objects being compared are a string.  The `ucKey` variable is already a string and `other.getUpperString()` also returns a string.  They are both guaranteed to be upper case so a case-insensitive comparison will be the fastest.

core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
```diff
@@ -260,8 +263,19 @@ public QueryImpl execute(PageContext pc, final SQL sql, int maxrows, int fetchsi
...
+			throw Caster.toPageException(qoqException);
+		}
+
+		// Debugging option to to log all QoQ that fall back on hsqldb in the datasource log
+		if (qoqException != null && Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.debug", "false"))) {
```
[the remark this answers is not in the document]
bdw429s: Fixed the same as above.

core/src/main/java/lucee/runtime/db/QoQ.java
bdw429s: @michaeloffner Java the Stream API and Lambdas are part of JDK 8.  Are you just saying the `Object[]::new` part is not Java 8 compat?  I can remove that to just use the `toArray()` method on the Stream.

bdw429s:
@michaeloffner I have addressed all of the items above in the latest commit.  Please review again.

core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
```diff
@@ -69,6 +70,23 @@
...
+		}
+		catch (PageException e) {
+			hsqldbDisable = false;
+		}
+		try {
+			hsqldbDebug = Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.debug", "false"));
+		}
+		catch (PageException e) {
+			hsqldbDebug = false;
+		}
```
michaeloffner, line 88: the try catch is not necessary in this case. "getSystemPropOrEnvVar" does not throw an error and 
 "Caster.toBooleanValue" can be changed as follow to avoid an exception
`hsqldbDisable = Caster.toBooleanValue(SystemUtil.getSystemPropOrEnvVar("lucee.qoq.hsqldb.disable", "false"),false);`
so set a default value for the method `toBooleanValue`

core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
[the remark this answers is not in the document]
michaeloffner: make sense and i agree, in other places env var are only read once, so a restart always is necessary.

core/src/main/java/lucee/runtime/db/QoQ.java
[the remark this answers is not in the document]
michaeloffner: It's fine as it is, i was just not sure about the naming convention.

core/src/main/java/lucee/runtime/db/QoQ.java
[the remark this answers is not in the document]
michaeloffner: yeah I understand your concern of modifying while looping, in that case i would do it as well as you did. But in that case i think we are safe to use entryset.

core/src/main/java/lucee/runtime/db/QoQ.java
michaeloffner: i'm always using different JVM to work and build, just be on the safe side, openJDK was not complaining, but Oracle did. I try to be as conservative as possible with new functionality, i only using it when i need it and if it does not break backward compatibility. But i think you know that game with CFML compatibility ;-).

core/src/main/java/lucee/runtime/db/QoQ.java
michaeloffner: ;-)

core/src/main/java/lucee/runtime/type/KeyImpl.java
michaeloffner: agree i did some testing and the result was clear (i only test with different instances of the string)
`	public static void main(String[] args) {
		int max = 1000000;
		int loops = 10;
		String strEQ1 = "bjguguhbjbjgftzfzvhvbkjbkjbjkbj";
		String strEQ2 = new String(strEQ1.getBytes(CharsetUtil.UTF8), CharsetUtil.UTF8); // make sure we do not have the same instance
		String strOD = "bjguguhbjbjgftzfzvhvbkjbkjbjkbi";

		// equal true
		double fastestEqualTrue = Double.MAX_VALUE;
		for (int y = 0; y < loops; y++) {
			double start = SystemUtil.millis();
			boolean b;
			for (int i = 0; i < max; i++) {
				b = strEQ1.equals(strEQ2);
			}
			double tmp = SystemUtil.millis() - start;
			if (fastestEqualTrue > tmp) fastestEqualTrue = tmp;
		}
		print.e("equal true:" + fastestEqualTrue);

		// equal true
		double fastestEqualFalse = Long.MAX_VALUE;
		for (int y = 0; y < loops; y++) {
			double start = SystemUtil.millis();
			boolean b;
			for (int i = 0; i < max; i++) {
				b = strEQ1.equals(strOD);
			}
			double tmp = SystemUtil.millis() - start;
			if (fastestEqualFalse > tmp) fastestEqualFalse = tmp;
		}
		print.e("equal false:" + fastestEqualFalse);

		// equalsIgnoreCase true
		double fastestEqualIgnoreCaseTrue = Long.MAX_VALUE;
		for (int y = 0; y < loops; y++) {
			double start = SystemUtil.millis();
			boolean b;
			for (int i = 0; i < max; i++) {
				b = strEQ1.equalsIgnoreCase(strEQ2);
			}
			double tmp = SystemUtil.millis() - start;
			if (fastestEqualIgnoreCaseTrue > tmp) fastestEqualIgnoreCaseTrue = tmp;
		}
		print.e("equalsIgnoreCase true:" + fastestEqualIgnoreCaseTrue);

		// equalsIgnoreCase false
		double fastestEqualIgnoreCaseFalse = Long.MAX_VALUE;
		for (int y = 0; y < loops; y++) {
			double start = SystemUtil.millis();
			boolean b;
			for (int i = 0; i < max; i++) {
				b = strEQ1.equalsIgnoreCase(strOD);
			}
			double tmp = SystemUtil.millis() - start;
			if (fastestEqualIgnoreCaseFalse > tmp) fastestEqualIgnoreCaseFalse = tmp;
		}
		print.e("equalsIgnoreCase false:" + fastestEqualIgnoreCaseFalse);
	}`

Result:
equal true:0.3223149999976158
equal false:0.3223139997571707
equalsIgnoreCase true:133.16265799989924
equalsIgnoreCase false:147.4729599999264

equal is MUCH faster

core/src/main/java/lucee/runtime/db/HSQLDBHandler.java
michaeloffner: this is a small change i will do it manually

bdw429s: That makes sense.  I was not aware using the overloaded method with the default value would avoid the checked exception, but in retrospect this makes sense.

bdw429s:
@michaeloffner Thanks again for the additional feedback and merge!!

PR #1030, opened 2020-09-11
use an ordered struct for getTagData attributes LDEV-3040

zspitzer:
Need to preserve the order of attributes from the XML definition

https://luceeserver.atlassian.net/browse/LDEV-3040


michaeloffner:
sorry we did already the same in meantime

PR #1036, opened 2020-09-15
add sessioncookie support for path attribute LDEV-3062  

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3062


zspitzer:
image

core/src/main/java/lucee/runtime/PageContextImpl.java
```diff
@@ -2688,15 +2689,21 @@ private void setClientCookies() {
 				if (!StringUtil.isEmpty(tmp, true)) domain = tmp.trim();
 				// samesite
 				samesite = data.getSamesite();
+				// path
+				String tmp2 = data.getPath();
+				if (StringUtil.isEmpty(tmp2, true))
+					path = "/";
```
michaeloffner: should in case "tmp2" is empty not simply keep what was set on line 2672?

zspitzer: Sure, done

PR #1043, opened 2020-09-18
LDEV-3076 fix for updatedefaultsecuritymanager in admin tag

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3076


core/src/main/java/lucee/runtime/security/SecurityManagerImpl.java
```diff
@@ -200,6 +200,8 @@ public static short toShortAccessValue(String accessValue) throws SecurityExcept
 		else if (accessValue.equals("8")) return VALUE_8;
 		else if (accessValue.equals("9")) return VALUE_9;
 		else if (accessValue.equals("10")) return VALUE_10;
+		else if (accessValue.equals("0")) return VALUE_NO;
+		else if (accessValue.equals("-1")) return VALUE_YES;
```
michaeloffner, line 204: seems very strange that -1 is an accepted input, in Lucee -1 is always equal to not set

PR #1044, opened 2020-09-20
add SAMESITE_EMPTY LDEV-2993

zspitzer:
- changed the default for samesite not to be NONE
- add samesite EMPTY (default)
- always set samesite none when defined

https://luceeserver.atlassian.net/browse/LDEV-2993


zspitzer:
My patch for the 5.3 branch will conflict with this due to overlap with https://github.com/lucee/Lucee/pull/1036

once that is merged, I will submit a PR for 5.3

PR #1049, opened 2020-09-24
Use bitwise operators to set Pattern flags

elpete:
Solves https://luceeserver.atlassian.net/browse/LDEV-3087


zspitzer:
apparently your fix doesn't work yet?
https://github.com/lucee/Lucee/pull/1050

elpete:
I'm not sure, then.  Sorry. :-/

PR #1052, opened 2020-10-01
add CFFeed support for columnMap LDEV-1770

zspitzer:
https://docs.lucee.org/reference/tags/feed.html#attribute-columnmap
https://luceeserver.atlassian.net/browse/LDEV-1770


zspitzer:
Test case is https://github.com/lucee/Lucee/pull/433

PR #1063, opened 2020-10-12
avoid unnecessary debug casting LDEV-3071

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3071


michaeloffner:
the _toString method does a little bit more than just convert to string
`private String _toString(long value) {
		if (value <= 0) return "0";
		return String.valueOf(value);
	}`

the `if (value <= 0) return "0";` part should remain.

zspitzer:
the only time they would be negative is if they overflowed?

zspitzer:
https://github.com/lucee/Lucee/commit/e4d71cff6c73d171c14baaa76e5df8b0e0ca1dca

PR #1064, opened 2020-10-12
corrected the action name in debugging.logs.cfm

cfmitrah:


zspitzer:
https://github.com/lucee/Lucee/pull/1064
https://github.com/lucee/Lucee/pull/1065

PR #1072, opened 2020-10-16
Add CFSCHEDULE userAgent LDEV-2999

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-2999

not quite working, it compiles but throws an error on usage


michaeloffner:
@zspitzer could you always use the same name character cases for user agent
we have setUseragent but then getUserAgent

zspitzer:
done!

I think I may of just copied the style from http.java and feed.java?

zspitzer:
problem is this changes the loader signature.....

PR #1087, opened 2020-10-24
LDEV-3112 fix

kabutotx:
Only return verArr[2] if verArr[1] under 2.
Fix for JDKs with only major version.

### DIFF
--- a/core/src/main/cfml/context/admin/overview.cfm
+++ b/core/src/main/cfml/context/admin/overview.cfm
@@ -773,7 +773,10 @@ Error Output --->
 <cfscript>
 	function getJavaVersion() {
 		var verArr=listToArray(server.java.version,'.');
-		if(verArr[1]>2) return verArr[1];
-		return verArr[2];
+		if(verArr[1]>2) {
+			return verArr[1];
+		} else {
+			return verArr[2];
+		}
 	}
 </cfscript>


core/src/main/cfml/context/admin/overview.cfm
```diff
@@ -773,7 +773,10 @@ Error Output --->
 <cfscript>
 	function getJavaVersion() {
 		var verArr=listToArray(server.java.version,'.');
-		if(verArr[1]>2) return verArr[1];
-		return verArr[2];
```
stillnet, line 777: Wasn't the 2nd return in the old code effectively an "else", similar to what the new code is doing? Sorry maybe it's too late in the evening for me but I'm not seeing how the new code fixes it. I would have expected an `elseif (ArrayLen(verArr) GT 1)` or something like that.

kabutotx:
Ugh, Yeh.  Guess I should have test with actual 16 JDK.  It returns "16-ea" which causes the error. I guess it is just an edge case since 16 is still in development. Version class would fix the problem but not for Java 8.

core/src/main/cfml/context/admin/overview.cfm
stillnet: @kabutotx  I think this will do it.

	function getJavaVersion() {
		var verArr=listToArray(server.java.version,'.');
		if(verArr[1]>2) {
			return verArr[1];
		} elseif (verArr.len() GT 1) {
			return verArr[2];
		} else {
		    return val(server.java.version);
		}
	}

image

PR #1089, opened 2020-10-24
LDEV-3112 Fix crash on versions of Java that to not match the expected format

stillnet:
Fix crash on versions of Java that to not match the expected format, for example "16-ea".


michaeloffner:
@stillnet thanks for your input

PR #1090, opened 2020-10-24
disable running debugging output after cfcontent LDEV-3074

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3074


zspitzer:
cflocation uses

`if (pageContext.getConfig().debug()) pageContext.getDebugger().setOutput(false);`

PR #1095, opened 2020-10-29
Fixed issue where MSSQL driver is not matching JTDS results LDEV-3102

dswitzer:
This patch resolves issues where the MSSQL driver did not already return the expected results when used with the "result" attribute on the query, which caused queries executed with the new Query() syntax to break.

https://luceeserver.atlassian.net/browse/LDEV-3102


dswitzer:
I realized there's a bug in the unit tests the way they're designed. I was trying to get Testbox to run the same tests over both JTDS and MSSQL, so I could verify behavior between the two drivers. However, the closures appear to get cached, so it ends up being the same test running on each iteration. 

I'll need to either re-do the tests, or you can just comment out the JTDS driver line for the time being.

dswitzer:
The unit tests were not working as expected, so I've fixed them. Fixing the unit tests, lead me to one of the tests that were failing for MSSQL, so I've fixed that now.

PR #1096, opened 2020-11-02
Fix exceptions not being raised when using the MSSQL driver LDEV-3127 

dswitzer:
This fixes the issue where using the MSSQL driver exceptions reported by SQL Server are ignored. This patch requires the LDEV-3102 fix, because there was overlap in the changes.

https://luceeserver.atlassian.net/browse/LDEV-3127
https://luceeserver.atlassian.net/browse/LDEV-3102


michaeloffner:
we need to closely test this change, this couild affect performance of any single query done. if that only applies to MSSQL, we should consider only to do it for MSSQL

dswitzer:
@michaeloffner 

I don't disagree. As I stated in the LDEV-3127, I kept thinking while working through the fixes that the code probably should be refactored to clean it up, but I was afraid a major refactoring might be more problematic in getting the fixes approved.

Also, I didn't do exhaustive performance testing by any means (and I'm sure you have better tools for it than I do), but I did not notice any issues. However, I was only really testing under MSSQL.

michaeloffner:
i do not like this line:
`if (DataSourceUtil.isMSSQLDriver(dc) && se.getMessage() != "The statement must be executed before any results can be obtained.") {`

this is relaying on the fact that this message will not change, what is a risk. there are many reason that message can change:
- support for locale based exception message
- rewording
- ...

dswitzer:
@michaeloffner: 

> 
> 
> i do not like this line:
> `if (DataSourceUtil.isMSSQLDriver(dc) && se.getMessage() != "The statement must be executed before any results can be obtained.") {`
> 
> this is relaying on the fact that this message will not change, what is a risk. there are many reason that message can change:
> 
>     * support for locale based exception message
> 
>     * rewording
> 
>     * ...

From the MSSQL source code, maybe `SQLServerException.getErrString("R_statementMustBeExecuted")` could be used instead:

https://github.com/microsoft/mssql-jdbc/blob/89bb744675941113a0ca9bcd7dc3c8f1310bb23c/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerStatement.java#L2205

michaeloffner:
maybe we can check the condition that caused the exception in the first place:
`if (!getNextResult(true) || null == resultSet) {`

dswitzer:
> 
> 
> maybe we can check the condition that caused the exception in the first place:
> `if (!getNextResult(true) || null == resultSet) {`

Of course if the logic in getGeneratedKeys() changes that throws the exception, the code could still break because the wrong verification code would be getting used.

Also, my memory is fuzzy, but I was thinking the `getNextResult(true)` ends up advancing the stream, so checking it first causes issues. I think that's the issue I was running into trying to manually check it.

michaeloffner:
code removes that check, why?
`if (keyName == null) {
							// single record struct
							if (recordcount > 0) {
								throw new ApplicationException("Attribute [keyColumn] is required when return type is set to Struct and more than one record is returned");
							}`

dswitzer:
@michaeloffner 

> 
> 
> code removes that check, why?
> `if (keyName == null) { // single record struct if (recordcount > 0) { throw new ApplicationException("Attribute [keyColumn] is required when return type is set to Struct and more than one record is returned"); }`

I apologize, I'm not seeing that in the changeset. Where in the original code was that?

I made these changes a while ago, so it's not fresh in my mind.

dswitzer:
@michaeloffner 

That line of code doesn't look like it's in the 5.3 branch, but is in the 6.0. So I doesn't look like I removed it, it just was not there.

michaeloffner:
@dswitzer here it is
https://github.com/lucee/Lucee/blob/5.3/core/src/main/java/lucee/runtime/type/QueryImpl.java#L599

michaeloffner:
most likely it does not show up in the online comparison because of the merge conflict, it was added here
https://github.com/lucee/Lucee/commit/866bd81f6413eda981dc2cc014dbb32540110d32

with version 5.3.8.132-SNAPSHOT

dswitzer:
@michaeloffner

That fix was added *after* my patch. It was added Dec 2020, where my patch was developed Nov 2, 2020.

https://github.com/lucee/Lucee/commits/ad8bdcc59a371784ebfd3d31ad6e3f93f8991389/core/src/main/java/lucee/runtime/type/QueryImpl.java

PR #455, opened 2018-04-19
improve admin text readability

zspitzer:
- remove italics
- remove light grey color
- use consistent font size

PR #1101, opened 2020-11-12
improve admin text readability (again)

zspitzer:
this improvement keeps getting reverted

https://github.com/lucee/Lucee/pull/455
https://github.com/lucee/Lucee/pull/659

change the comments style in admin to be more readable

- larger font 
- darker text
- no italics


michaeloffner:
i do not see any reversion
https://github.com/lucee/Lucee/commits/5.3/core/src/main/cfml/context/res/css/admin.css

PR #1108, opened 2020-11-26
https://luceeserver.atlassian.net/browse/LDEV-3153 - added StringSort…

mircea-botez:
… function


mircea-botez:
i will be creating another pr.

PR #1109, opened 2020-12-01
Added a testcase for LDEV-3167

cfmitrah:


zspitzer:
this test case would be better written to just expect at least x items in the available methods list, otherwise it will need to be continually revised as methods are added

[append,clear,copy,count,delete,duplicate,each,every,filter,find,findKey,findValue,insert,isEmpty,keyArray,keyExists,keyList,keyTranslate,len,map,reduce,some,sort,update]

PR #1118, opened 2020-12-14
Changed the test case and added fix for LDEV-0946

cfmitrah:


core/src/main/java/lucee/runtime/ComponentImpl.java
```diff
@@ -1596,6 +1596,8 @@ private static void metaUDFs(PageContext pc, ComponentImpl comp, Struct sct, int
 		if (arr.size() != 0) {
 			Collections.sort(arr, new ComparatorImpl());
 			sct.set(KeyConstants._functions, arr);
+		}else{
```
jbampton, line 1599:
```suggestion
		} else {
```

jbampton (changes requested):

core/src/main/java/lucee/runtime/ComponentImpl.java
michaeloffner: @jbampton manual format changes are not necessary, we auto format all code
https://github.com/lucee/Lucee/tree/5.3/core/.settings

PR #1125, opened 2020-12-15
return debug logs as json via url param

zspitzer:
add support for format=json for the debug log lists and log detail page

the idea behind this is, an IDE extension could access this as a JSON api

and use it to show the request code path, show execution times, flag implicit variable problems etc

### DIFF
--- a/core/src/main/cfml/context/admin/debugging.logs.cfm
+++ b/core/src/main/cfml/context/admin/debugging.logs.cfm
@@ -147,13 +147,19 @@ Redirtect to entry --->
     
     <cfreturn (time/1000000)&" ms">
 </cffunction> 
-    
+<cfscript>
+	param name="url.action2" default="list";
+	param name="url.format" default="";
 
-
-<cfparam name="url.action2" default="list">
-
-<cfif url.action2 EQ "list">
-	<cfinclude template="debugging.logs.list.cfm">
-<cfelse>
-	<cfinclude template="debugging.logs.detail.cfm">
-</cfif>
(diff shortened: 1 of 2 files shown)


core/src/main/cfml/context/admin/debugging.logs.cfm
```diff
@@ -147,13 +147,19 @@ Redirtect to entry --->
...
+	param name="url.format" default="";
 
-<cfparam name="url.action2" default="list">
-
-<cfif url.action2 EQ "list">
-	<cfinclude template="debugging.logs.list.cfm">
-<cfelse>
-	<cfinclude template="debugging.logs.detail.cfm">
-</cfif>
+	if (url.action2 EQ "list"){
```
jbampton:
```suggestion
	if (url.action2 EQ "list") {
```

core/src/main/cfml/context/admin/debugging.logs.cfm
```diff
@@ -147,13 +147,19 @@ Redirtect to entry --->
...
+		if (url.format eq "json"){
```
jbampton:
```suggestion
		if (url.format eq "json") {
```

core/src/main/cfml/context/admin/debugging.logs.detail.cfm
```diff
@@ -52,6 +52,13 @@
 			log=el;
 		}
 	}
+	
+	if (url.format eq "json"){
```
jbampton:
```suggestion
	if (url.format eq "json") {
```

PR #1131, opened 2020-12-18
add stringEvery LDEV-3180 

mircea-botez:
https://luceeserver.atlassian.net/browse/LDEV-3180


michaeloffner:

PR #1134, opened 2020-12-18
add StringReduce LDEV-3160

mircea-botez:
https://luceeserver.atlassian.net/browse/LDEV-3160

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/closure/Reduce.java
+++ b/core/src/main/java/lucee/runtime/functions/closure/Reduce.java
@@ -55,7 +55,11 @@ public class Reduce extends BIF implements ClosureFunc {
 		return _call(pc, obj, udf, initalValue, TYPE_UNDEFINED);
 	}
 
-	public static Object _call(PageContext pc, Object obj, UDF udf, Object initalValue, short type) throws PageException {
+	public static Object call(PageContext pc, Object obj, UDF udf, Object initalValue, short type) throws PageException {
+		return _call(pc, obj, udf, initalValue, type);
+	}
+
+	private static Object _call(PageContext pc, Object obj, UDF udf, Object initalValue, short type) throws PageException {
 
 		Object value;
 
--- a/core/src/main/java/lucee/runtime/functions/string/StringReduce.java
+++ b/core/src/main/java/lucee/runtime/functions/string/StringReduce.java
@@ -4,12 +4,13 @@ import lucee.runtime.PageContext;
 import lucee.runtime.exp.FunctionException;
 import lucee.runtime.exp.PageException;
 import lucee.runtime.ext.function.BIF;
+import lucee.runtime.ext.function.Function;
 import lucee.runtime.functions.closure.Reduce;
 import lucee.runtime.op.Caster;
 import lucee.runtime.type.UDF;
 import lucee.runtime.type.util.StringListData;
 
-public class StringReduce extends BIF {
+public class StringReduce extends BIF implements Function {
 	private static final long serialVersionUID = -2153555241217815037L;
 
 	@Override
--- a/core/src/main/java/lucee/runtime/functions/string/StringReduce.java
+++ b/core/src/main/java/lucee/runtime/functions/string/StringReduce.java
@@ -1,0 +1,31 @@
+package lucee.runtime.functions.string;
+
+import lucee.runtime.PageContext;
+import lucee.runtime.exp.FunctionException;
+import lucee.runtime.exp.PageException;
+import lucee.runtime.ext.function.BIF;
+import lucee.runtime.functions.closure.Reduce;
+import lucee.runtime.op.Caster;
+import lucee.runtime.type.UDF;
+import lucee.runtime.type.util.StringListData;
+
+public class StringReduce extends BIF {
+	private static final long serialVersionUID = -2153555241217815037L;
+
+	@Override
+	public Object invoke(PageContext pc, Object[] args) throws PageException {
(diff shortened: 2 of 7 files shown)


core/src/main/java/lucee/runtime/functions/closure/Reduce.java
```diff
@@ -55,7 +55,11 @@ public static Object call(PageContext pc, Object obj, UDF udf, Object initalValu
 		return _call(pc, obj, udf, initalValue, TYPE_UNDEFINED);
 	}
 
-	public static Object _call(PageContext pc, Object obj, UDF udf, Object initalValue, short type) throws PageException {
+	public static Object call(PageContext pc, Object obj, UDF udf, Object initalValue, short type) throws PageException {
+		return _call(pc, obj, udf, initalValue, type);
```
michaeloffner, line 59: why that change?

core/src/main/java/lucee/runtime/functions/string/StringReduce.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class StringReduce extends BIF {
+	private static final long serialVersionUID = -2153555241217815037L;
+
+	@Override
+	public Object invoke(PageContext pc, Object[] args) throws PageException {
+		if (args.length < 2) {
+			throw new FunctionException(pc, "StringSome", 2, 3, args.length);
+		}
+		return call(pc, Caster.toString(args[0]), Caster.toFunction(args[1]), args[2]);
```
michaeloffner: that will fail when we only have 2 arguments in the request, because line 3 is expecting 3 arguments and line 17 only handles less that 2.
i mostly do it the other way arround, so
if(args.length==2) return call(pc, Caster.toString(args[0]), Caster.toFunction(args[1]));
else if(args.length==3) return call(pc, Caster.toString(args[0]), Caster.toFunction(args[1]), args[2]);
else throw new FunctionException(pc, "StringSome", 2, 3, args.length);

michaeloffner: correction:
if(args.length==2) return call(pc, Caster.toString(args[0]), Caster.toFunction(args[1]),null);
else if(args.length==3) return call(pc, Caster.toString(args[0]), Caster.toFunction(args[1]), args[2]);
else throw new FunctionException(pc, "StringSome", 2, 3, args.length);

core/src/main/java/lucee/runtime/functions/string/StringReduce.java
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	public static Object call(PageContext pc, String inputString, Object value, Object initialValue) throws PageException {
+		StringListData stringList = new StringListData(inputString, "", false, false);
```
michaeloffner: can you use the same name pattern as we have for ArrayReduce and StructReduce so
`public static Object call(PageContext pc, String str, UDF udf, Object initValue)`

core/src/main/java/resource/fld/core-base.fld
```diff
@@ -13850,4 +13850,35 @@ You can find a list of all available timezones in the Lucee administrator (Setti
...
+    
+    <function>
+		<name>StringReduce</name>
+		<member-name>reduce</member-name>
+		<class>lucee.runtime.functions.string.StringReduce</class>
+		<description>Iterates over every element of the string and calls the closure to work on the elements of the string. This function will reduce the string to a single value and will return the value.</description>
+    	<argument>
+			<name>string</name>
+			<type>string</type>
+			<alias>inputString, string</alias>
```
michaeloffner: alias for "string" is not necessary, it is already the name
add "object" as alias, as structReduce has

core/src/main/java/resource/fld/core-base.fld
```diff
@@ -13850,4 +13850,35 @@ You can find a list of all available timezones in the Lucee administrator (Setti
...
+			<required>Yes</required>
+			<description> Closure or a function reference that will be called for each of the iteration.</description>
+		</argument>
+		<argument>
+			<name>initialVal</name>
+			<type>any</type>
+			<alias>initialVal</alias>
+			<required>No</required>
+			<description> Initial value that will be used for the reduce operation.</description>
+		</argument>
```
michaeloffner: follow here also the structReduce example that looks like this
```
<argument>
			<name>closure</name>
			<alias>function,callback,udf</alias>
			<type>function</type>
			<required>Yes</required>
			<description>function/closure that implements the following constructor [function(any result, string key, object value[, struct struct]):any].</description>
		</argument>
```

core/src/main/java/lucee/runtime/functions/string/StringReduce.java
```diff
@@ -0,0 +1,31 @@
...
+		if (value instanceof UDF) {
+			return Reduce.call(pc, (Object) stringList, (UDF) value, initialValue);
```
michaeloffner: check for UDF is not necessary, define the type in the function

core/src/main/java/resource/fld/core-base.fld
michaeloffner: the comment above is for the argument "callback"

core/src/main/java/lucee/runtime/functions/closure/Reduce.java
mircea-botez: just to be consistent, everything starting with _ is private

michaeloffner:
duplicate already merge here https://github.com/lucee/Lucee/pull/1340

PR #1135, opened 2020-12-22
CF / Lucee incompatibility for bit values and cfupdate LDEV-2606

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2606


michaeloffner:
We cannot do that `if(str.equalsIgnoreCase("on")) str = "true";` this is FAR to expensive, the function toBooleanValue is called a LOT and this is an absolute edge case. On contrary we should consider to remove support for "yes" or "no".

PR #1145, opened 2020-12-28
Added a fix and testcase for function week() in LDEV-3199

cfmitrah:


michaeloffner:
i did a different more safe approach, but i adapted the test case manually

PR #1151, opened 2020-12-30
fix for LDEV-3199

kabutotx:
Fix for week() at end of December on leap year.
Should only be 54 weeks on leap years starting on Sat Jan 1.  This should now give the same results as ACF.


zspitzer:
this duplicates https://github.com/lucee/Lucee/pull/1145 exactly :) (which also includes a testcase)

PR #1156, opened 2021-01-18
don't dump html response on extension download error LDEV-3226

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3226


zspitzer:
- this logs to `deploy.log` which is the standard location for extension stuff, but `printError` also logs out to `application.log` so it's creating duplicated logs
- ideally the the encodeForHtml() should happen in print error, but I'm not sure if that might break something else

PR #1165, opened 2021-02-02
ArrayToStruct add valueAsKey  LDEV-3171

zspitzer:
also changed the struct to be ordered

https://luceeserver.atlassian.net/browse/LDEV-3171

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/arrays/ArrayToStruct.java
+++ b/core/src/main/java/lucee/runtime/functions/arrays/ArrayToStruct.java
@@ -35,12 +35,15 @@ public final class ArrayToStruct extends BIF {
 
 	private static final long serialVersionUID = 2050803318757965798L;
 
-	public static Struct call(PageContext pc, Array arr) throws PageException {
-		Struct sct = new StructImpl();
+	public static Struct call(PageContext pc, Array arr, boolean valueAsKey) throws PageException {
+		Struct sct = new StructImpl(Struct.TYPE_LINKED);
 		int[] keys = arr.intKeys();
 		for (int i = 0; i < keys.length; i++) {
 			int key = keys[i];
-			sct.set(KeyImpl.toKey(key + ""), arr.getE(key));
+			if (valueAsKey)
+				sct.set(KeyImpl.toKey(arr.getE(key) + ""), key);
+			else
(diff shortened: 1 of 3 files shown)


core/src/main/java/lucee/runtime/functions/arrays/ArrayToStruct.java
```diff
@@ -35,20 +35,24 @@
...
-	public static Struct call(PageContext pc, Array arr) throws PageException {
-		Struct sct = new StructImpl();
+	public static Struct call(PageContext pc, Array arr, boolean valueAsKey) throws PageException {
+		Struct sct = new StructImpl(Struct.TYPE_LINKED);
 		int[] keys = arr.intKeys();
 		for (int i = 0; i < keys.length; i++) {
 			int key = keys[i];
-			sct.set(KeyImpl.toKey(key + ""), arr.getE(key));
+			if (valueAsKey)
+				sct.set(KeyImpl.toKey(arr.getE(key) + ""), key);
```
michaeloffner: here you need to do Caster.toKey(arr.getE(key)) instead, for the int it is fine to do `key+""`, but the value set can be everything, an array a component, whatever and if it is a complex value we need an exception.

core/src/main/java/lucee/runtime/functions/arrays/ArrayToStruct.java
```diff
@@ -35,20 +35,24 @@
 
 	private static final long serialVersionUID = 2050803318757965798L;
 
-	public static Struct call(PageContext pc, Array arr) throws PageException {
-		Struct sct = new StructImpl();
+	public static Struct call(PageContext pc, Array arr, boolean valueAsKey) throws PageException {
+		Struct sct = new StructImpl(Struct.TYPE_LINKED);
 		int[] keys = arr.intKeys();
```
michaeloffner: when touching that code, i would also improve it, instead of getting the key array and then get every single value separately, we should do here `arr.getEntryIterator()` and then do the following loop below
Entry a;
while(it.hasNext()) {
   e=it.next();

zspitzer: done

PR #1169, opened 2021-02-03
allow least min value on negative index array LDEV-3258

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3258


cfmitrah:
@zspitzer  I’ve added a correction in the PR. Also  updated and enabled the testcase in the linked issue LDEV-1836

PR #1178, opened 2021-02-07
cfldap timeout LDEV-1860

zspitzer:
- correct documentation to be in milliseconds (not seconds)
- always set timeout for all LDAP operations

https://luceeserver.atlassian.net/browse/LDEV-1860


zspitzer:
related, Lucee still bundles very old LDAP libraries which are now built in

https://github.com/lucee/Lucee/pull/1179

PR #1185, opened 2021-02-10
don't log out exception message three times, do it only twice

zspitzer:
it's already logged twice

https://luceeserver.atlassian.net/browse/LDEV-2781


michaeloffner:
adapted the change manually, because of merge conflicts
https://github.com/lucee/Lucee/commit/ed06801b77ac38eb6093d06b9ec8d68afd0a2d4b

PR #1188, opened 2021-02-12
split the function or key values into detail in the exception LDEV-2416

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2416


cfmitrah:
Solved by https://github.com/lucee/Lucee/pull/1242

PR #1190, opened 2021-02-12
Add more detailed Extension logging LDEV-3277

zspitzer:
log out to deloy.log

- update extension
- remove extension
- install bundle

https://luceeserver.atlassian.net/browse/LDEV-3277
https://luceeserver.atlassian.net/browse/LDEV-3032


zspitzer:
improved logging when a extension mainfest entry like resource doesn't get updated

previously it always logged out that it had updated, it now logs an error when the entry gets skipped

https://luceeserver.atlassian.net/browse/LDEV-3286

PR #1193, opened 2021-02-16
Added a fix and test case for isNumeric() string values

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3275


michaeloffner:
adapted the change manually, but not the change.

PR #1202, opened 2021-02-23
Added a new function directoryInfo() and action info in cfdirectory

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3294


zspitzer:
rather than a try catch, how about checking if the resource type is a `FileResource`, like with `cfexecute`?

https://github.com/lucee/Lucee/pull/1199

same for fileCreated https://github.com/lucee/Lucee/pull/1182

cfmitrah:
I closed the above PR due to conflict and created a new PR (https://github.com/lucee/Lucee/pull/1333)

PR #1205, opened 2021-02-24
rename unique column to uniqueTask LDEV-2755

zspitzer:
unique is a reserved word in sql and prevents using this with QoQ

https://luceeserver.atlassian.net/browse/LDEV-2755


michaeloffner:
we cannot simply rename that column, that will break a lot of existing code including the Lucee admin, that expect that column name to exist. Also when i run that code example from the ticket, i don't get that exception, more details in the ticket.

PR #1207, opened 2021-02-26
add QueryRenameColumn

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3208


core/src/main/java/resource/fld/core-base.fld
```diff
@@ -9930,6 +9930,38 @@ You can find a list of all available timezones in the Lucee administrator (Setti
...
 	</function>
 
+	<!-- queryRenameColumn -->
+	<function>
+		<name>queryRenameColumn</name>
+		<class>lucee.runtime.functions.query.QueryRenameColumn</class>
+		<keywords>query</keywords>
+		<member-name>renameColumn</member-name>
+		<description>Rename a column</description>
+		<introduced>6.0.0</introduced>
```
michaeloffner, line 9940: in other place (for example Manifest.mf) we call this "since"

core/src/main/java/resource/fld/core-base.fld
```diff
@@ -9930,6 +9930,38 @@ You can find a list of all available timezones in the Lucee administrator (Setti
...
+		<argument>
+			<name>query</name>
+			<type>query</type>
+			<required>Yes</required>
+			<description>Name of a query object.</description>
```
michaeloffner, line 9945: it is not the name of the query,it is the query itself. it could be for example queryRenameColumn(queryExecute(...),...)

PR #1255, opened 2021-03-27
Log native QoQ exceptions as a debug exception LDEV-3400

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3400

### DIFF
--- a/core/src/main/java/lucee/runtime/type/QueryImpl.java
+++ b/core/src/main/java/lucee/runtime/type/QueryImpl.java
@@ -1120,9 +1120,9 @@ public class QueryImpl implements Query, Objects, QueryResult {
 		StringBuilder sb = new StringBuilder();
 		for (int i = 0; i < columnNames.length; i++) {
 			if (i > 0) sb.append(delim);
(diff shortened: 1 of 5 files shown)


core/src/main/java/lucee/runtime/type/QueryImpl.java
```diff
@@ -1110,18 +1111,23 @@ public int getCurrentrow(int pid) {
...
 	public String getColumnlist(boolean upperCase) {
+		return getColumnlist(upperCase, ",");
+	}
+
+	public String getColumnlist(boolean upperCase, String delim) {
 		StringBuilder sb = new StringBuilder();
 		for (int i = 0; i < columnNames.length; i++) {
-			if (i > 0) sb.append(',');
+			if (i > 0) sb.append(delim);
 			sb.append(upperCase ? columnNames[i].getUpperString() : columnNames[i].getString());
```
isapir: It would be much more performant to UPPERCASE the result once at the end, no?

zspitzer: yeah, nice catch :)

PR #1263, opened 2021-04-04
Improve extension logging 6.0 LDEV-3032

zspitzer:
- tweak language, update is now installing
- changed default deploy log level to INFO
- changed all the verbose install entries to DEBUG
- log when an extension manifest entry install fails due to any problem
- log individual file installs under "real ext name" rather than generically as "extension"

https://luceeserver.atlassian.net/browse/LDEV-3032


zspitzer:
https://github.com/lucee/Lucee/pull/1190

michaeloffner:
we should NOT change the default log level, unless there is really a good reason for it

michaeloffner:
url,path,url is not necessary to get an extension, id and version can be sufficient. Actually id on it's own should be sufficient, when you define no version, it simply should pick the latest version.

zspitzer:
my rationale for changing the log level to info was to log out a single log entry for all installs / upgrades, 

as I moved all the detailed logging to DEBUG level. logging out upgrades / installs as warnings or errors doesn't quite feel right

I'll change it back to ERROR then

michaeloffner:
@cfmitrah the underlaying ticket is already set to resolved. Does this PR still address an existing issue or is it already obsolete? If it still is an issue, please create a new ticket for the issue and a new PR.

PR #1265, opened 2021-04-06
LDEV-3427 better udf debug logging

zspitzer:
more of a POC than anything

UDFs don't get logged in template execution times, only cfc function methods

https://luceeserver.atlassian.net/browse/LDEV-3427

image


michaeloffner:
@zspitzer why only doing this for function call without named values? also we should avoid making an addional method for this, to keep the overhead because of this to a minimum.

zspitzer:
closing see linked newer PR

PR #1270, opened 2021-04-09
len() tests for arrays, structs & query objects

JamoCA:
According to [documentation](https://docs.lucee.org/reference/functions/len.html), `len()` returns length for "string, array,struct,query ...".  Arrays & structs seem to work, but query objects appear to throw an error.
```
The function [len] does not exist in the Object, only the following functions are available: [addColumn,addRow,columnArray,columnCount,columnData,columnExists,columnList,currentRow,deleteColumn,deleteRow,duplicate,each,every,filter,getCell,getCellByIndex,getRow,keyExists,map,recordCount,reduce,rowByIndex,rowData,rowDataByIndex,setCell,slice,some,sort,valueArray,valueList]
```

### DIFF
--- a/test/jira/Jira2998.cfc
+++ b/test/jira/Jira2998.cfc
@@ -59,6 +59,17 @@ component extends="org.lucee.cfml.test.LuceeTestCase"	{
 	}
 	public void function testLen(){
 		assertEquals(3,"abc".len());
+		assertEquals(0, [].len());
+		assertEquals(0, {}.len());
+		assertEquals(0, {}.len());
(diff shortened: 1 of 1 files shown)


zspitzer:
`query.len()` doesn't exist, `query.recordcount()` does

https://github.com/lucee/Lucee/commit/4d5df06c33a62cc55841b9a16da80d0e87598695

I fixed the docs
https://github.com/lucee/Lucee/commit/33a524b9fb774191acf7a333d9706b3c25314dd5

PR #1278, opened 2021-04-12
FileAppend does not support file object LDEV-3424 

mircea-botez:
…s also an object


zspitzer:
closed, see https://github.com/lucee/Lucee/pull/1300

PR #1285, opened 2021-04-15
Enhancement: Fix to return new directory name instead of null/void for DirectoryRename().

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3453


zspitzer:
looks good, 

needs a test
https://github.com/lucee/Lucee/blob/6.0/test/functions/DirectoryRename.cfc

need to update the .fld too
https://github.com/lucee/Lucee/blob/6.0/core/src/main/java/resource/fld/core-base.fld#L3610

cfmitrah:
Done @zspitzer

PR #1291, opened 2021-04-20
Add warning message for overwrite the existing events in admin for LDEV-3464

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3464


michaeloffner:
@cfmitrah please check if that PR still makes sense

PR #1293, opened 2021-04-21
Added a Fix to retrieve the extension dropdown back in LDEV-3468

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3468


cfmitrah:
@zspitzer added another output instead remove encode for html

PR #1296, opened 2021-04-22
support custom struct for additional args for updateDatasource() in Administrator.cfc

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3469

### DIFF
--- a/core/src/main/java/resource/component/org/lucee/cfml/Administrator.cfc
+++ b/core/src/main/java/resource/component/org/lucee/cfml/Administrator.cfc
@@ -388,10 +388,10 @@ component {
 
 		var driver=createObject("component", drivernames[ arguments.type ]);
 		// var custom=structNew();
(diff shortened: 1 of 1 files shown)


core/src/main/java/resource/component/org/lucee/cfml/Administrator.cfc
```diff
@@ -386,18 +387,17 @@ component {
 		driverNames=ComponentListPackageAsStruct("dbdriver",driverNames);
 
 		var driver=createObject("component", drivernames[ arguments.type ]);
-		var custom=structNew();
+		// var custom=structNew();
 		loop collection="#arguments#" item="key"{
```
zspitzer: while we are touching this, that should be `local.key `

zspitzer (changes requested):

zspitzer (approved):

PR #1309, opened 2021-05-02
LDEV-4441 don't deploy admin when not enabled

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4441


zspitzer:
the problem with this approach is that this all this context deploy code is only called when the `/WEB-INF/` directory doesn't exist

PR #1314, opened 2021-05-04
Added a warning message in admin update while downgrade from 6 to 5

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3482


zspitzer:
Can you change 

_Please be aware when you update to Lucee 5, you will loose all configuration changes made since updating to Lucee 6._

to 

_Lucee 6 uses a different configuration storage mechanism (.json) than Lucee 5 (.xml). If you downgrade to Lucee 5.x, any configuration changes you have made in Lucee 6 will not be available. Lucee will revert back to the configuration before you upgraded to Lucee 6._

- this needs to be localizable (so move to language files)

cfmitrah:
@zspitzer I moved the text to the language file

michaeloffner:
code has changed a lot since this PR was made, so it no longer applies. Sorry for that @cfmitrah

PR #1317, opened 2021-05-07
add a check that method is defined for cfinvoke with a webservice LDEV-3493

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3493


cfmitrah:
@zspitzer same changes needed in doComponent() also.

PR #1325, opened 2021-05-13
Moved the whole debugging.template pages into debugging.setting page for LDEV-3489

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3489


michaeloffner:
@cfmitrah please resolve the merge conflict if the PR still applies for the current version and also makes sense

PR #1326, opened 2021-05-14
Cfdirectory mode is now applied to new parent dirs with createPath=true LDEV-3506

mircea-botez:


zspitzer:
the new test case for cfdirectory needs mode checking, which we need from https://github.com/lucee/Lucee/pull/1202

PR #1327, opened 2021-05-16
Fix spelling mistake

dariusj18:


zspitzer:
can you file it again against the 6.0 branch?

while you are there, can you add a comma after scope?

PR #1330, opened 2021-05-18
query returnType array should struct return row as an ordered struct LDEV-3511

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3511


zspitzer:
@cfmitrah can you add the test case as supplied in the issue? 

I checked and we don't have any coverage for this return type

cfmitrah:
@zspitzer done.

PR #1351, opened 2021-06-08
Added a fix for hash() and tobase64() member function to throw correct error with struct and array

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3556


cfmitrah:
@zspitzer I added a testcase

PR #1357, opened 2021-06-16
Fix single quotes and wddx LDEV-3573

bdw429s:
WDDX doesn't escape single ticks in struct keys creating invalid XML


bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-3573

cc/ @zspitzer

bdw429s:
@zspitzer Is there a way to re-run the job on this pull?  I think it was failing for an unrelated reason at the time I sent it and now the logs are too old to view and I can't find a way to see if it's passing now.

PR #1368, opened 2021-07-01
add test for zip executable LDEV-2117

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-2117


zspitzer:
seeing some strange results

- directoryList isn't returning the mode for either of the test files (one is executable, one isn't)
- the second file is also being unzipped with an execute flag

PR #1379, opened 2021-07-09
Added a correction to getComponentMetaData() return static Properties in staticFunctions key

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3362


michaeloffner:
manually adapted the change

PR #1381, opened 2021-07-13
Added a fix to NPE in admin when file field is not file for LDEV-3611

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3611


michaeloffner:
this is already fixed, see the tickets for details

PR #1391, opened 2021-07-21
Added a testcase calling component in a thread for LDEV-3627

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3627


zspitzer:
different testcase was added by Micha, so I'll close this PR

PR #1395, opened 2021-07-26
fix: Don't use the Lucee DatasourceManager for ORM connections - LDEV-1564

michaelborn:
Fixes LDEV-1564 by removing the use of the Lucee DatasourceManager for ORM connections. 

See this commit on the Hibernate extension - it's clear that the ORM extension contains its own session / connection manager.

https://github.com/lucee/extension-hibernate/commit/fe8c7f314f3ca98cd4d8d468399414eac24b8718

https://luceeserver.atlassian.net/browse/LDEV-1564


michaeloffner:
this is not about using the Lucee DatasourceManager or not, this is about registering session for transactions, you cannot simple disable this. You disable transactions completely. In addition we will not do that kind of change in Lucee itself (specially not 5.3). this brakes backward compatibility for older hibernate versions as well.

PR #1397, opened 2021-07-26
LDEV-3615 QoQ mishandles null and boolean column aliases

bdw429s:


bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-3615

PR #1398, opened 2021-07-27
LDEV-3640 QoQ needs to preserve nulls internally regardless of full null support

bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-3640


bdw429s:
I made a few subtle, but important changes here
* stop detecting null support and returning strings internally.  Just use nulls exclusively inside the query object-- they'll be converted later when the query is used if necessary 
* Make sure I pass in a default value of `null` any time I access the value of a cell to prevent the null-default of an empty string kicking in
* Ensure all cell access is via `QueryColumn.get()`, which is the only way to get a a raw cell value and NOT have the nulls changed to empty strings.

PR #1400, opened 2021-07-30
Added a fix to support to passing queryString in _InternalRequest() for LDEV-3617

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3617


michaeloffner:
sorry to decline this, i did use a different aproach that allowed to support this for forms as well at the same time.
https://github.com/lucee/Lucee/commit/7cf95d68a0cd567475d36d9735eeca8582bff289

PR #1405, opened 2021-08-09
Added a fix to catches the classCastException in duplicator for LDEV-3658 & LDEV-3752

cfmitrah:
### Fixes
- https://luceeserver.atlassian.net/browse/LDEV-3658
- https://luceeserver.atlassian.net/browse/LDEV-3752

### DIFF
--- a/core/src/main/java/lucee/runtime/op/Duplicator.java
+++ b/core/src/main/java/lucee/runtime/op/Duplicator.java
@@ -159,6 +159,9 @@ public final class Duplicator {
 		catch (ClassException e) {
 			newList = new ArrayList();
 		}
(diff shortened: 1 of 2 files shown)


bdw429s:
@cfmitrah @michaeloffner @zspitzer This pull is not the correct fix for the issue and IMO should not be merged.  The real bug is actually in the `ClassUtil.java` class which returns an exception object instead of throwing it.  What you've attempted to do is fix a secondary downstream error.  Please see this pull of mine which contains the proper fix.

https://github.com/lucee/Lucee/pull/1508

With that pull, you won't have the class cast exceptions in the first place.

PR #1412, opened 2021-08-18
Added a fix to getMetadata() add final key for final component in LDEV-3670

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3670


cfmitrah:
The same PR for the abstract component: https://github.com/lucee/Lucee/pull/1361

PR #1417, opened 2021-08-24
Added a fix for cfmail crashes on email addresses with trailing commas in LDEV-3687

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3687


zspitzer:
Can you add some test cases?

cfmitrah:
Test case added @zspitzer

PR #1419, opened 2021-08-27
add cfinputclassic encodeValue=false

zspitzer:


bdw429s:
@zspitzer It does seem appropriate for CFInput to automatically encode values.  Perhaps instead of a flag to turn encoding off, there should be a setting to change it to use the ESAPI encoders.  Also note, this will make the CFAjax extension have a dependency of the ESAPI extension, which seems very reasonable to me, but I'm not sure if Lucee has a concept of extension dependencies right.

zspitzer:
For Lucee 6, all the older encoders like htmleditformat etc, should use ESAPI and be secure by default.

We can't change core behaviour in 5.3, but 6 should be secure by default again.

I think this security stuff is core language, but we need to be able to update it between releases

PR #1425, opened 2021-09-07
Added a fix to reMatch/reMatchNoCase regression with empty string for LDEV-3703

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3703


test/tickets/LDEV3703.cfc
```diff
@@ -0,0 +1,17 @@
+component extends = "org.lucee.cfml.test.LuceeTestCase" {
```
zspitzer: can you add labels="regex"

cfmitrah: @zspitzer I added labels

cfmitrah:
@zspitzer I added labels

PR #1427, opened 2021-09-09
LDEV-3711

bdw429s:
Preserve the cause in DatabaseExceptions which can contain valuable nested exceptions from the JDBC driver.

### DIFF
--- a/core/src/main/java/lucee/runtime/exp/DatabaseException.java
+++ b/core/src/main/java/lucee/runtime/exp/DatabaseException.java
@@ -49,7 +49,7 @@ public final class DatabaseException extends PageExceptionImpl {
 
 		set(sqle);
 		set(dc);
-		initCause( sqle.getCause() );
+		initCause( sqle );
 	}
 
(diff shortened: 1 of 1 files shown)


bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-3711

michaeloffner:
this is wrong that way, in my opinion it should be
`initCause( sqle );`
the SQLException itself already is the cause, what you do is the cause of the cause. The SQL exception itself may has no cause.
But otherwise i agree with the PR a 100%

bdw429s:
@micstriit Problem is I tried that originally and it did not work for some reason.  I had this code:
```js
if( sqle.getCause().getCause() != null ) {
  initCause( sqle.getCause().getCause() );
}
```
and the if statement was not entered for the MySQL example I discussed in the ticket.  When I removed the if and set the top cause, the correct behavior was observed in my tests resulting int he correct stack trace that I put in the ticket.  

So what you're saying makes sense, but it doesn't seem to be correct.  The code I sent in the pull is what actually worked in real life.

bdw429s:
@micstriit I have updated the pull.

PR #1432, opened 2021-09-17
Added a fix to deserializeJSON() throw error for empty String in LDEV-3413

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3413


bdw429s:
@cfmitrah Please make sure however that this example still works:

```
deserializeJSON( '""' )
```

michaeloffner:
@bdw429s it does not affect it

PR #1439, opened 2021-09-29
LDEV-3854 - Performance/page spool lock contention

xdecock:
Rewrote the PageSpool code to rely on java8 atomic operations, this allows to remove the synchronized that were all over the place, reducing contention.

Furthermore, the cleanup has been implemented with an high / low watermark and compare and swap, to avoid removing a page that has just been regenerated.

https://luceeserver.atlassian.net/browse/LDEV-3854
(was https://luceeserver.atlassian.net/browse/LDEV-2904)

https://dev.lucee.org/t/blocked-requests-each-minute-on-5-3-8-x-under-load/8943


zspitzer:
@xdecock Lucee targets 11 but still supports java 8, 

https://github.com/ben-manes/caffeine#download 

says use the 2x branch for java 8 support

xdecock:
I'm in the process of rebuilding our "in house" build on 5.4 I'll do some repass in the coming week to make sure we're up to date + make sure the system handles the load without issue.

Thanks for the follow-up.

michaeloffner:
for details please see the comments in the ticket. thanks for the PR anyway, great work!

PR #1444, opened 2021-09-30
Added a fix to GetTagData() & GetFunctionData() to remove extra tabs in description for LDEV-1811

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1811


michaeloffner:
i would prefer to support OPTIONAL Markdown support for docs coming from fld/tld files, if we use some external kib we could also add some functions to lucee like markdownToHTML

zspitzer:
if we move in that direction, we should write a script to re-import all the reformatted MD content in lucee-docs back into the .tld and .fld's

PR #1447, opened 2021-10-11
Added a function ArrayRemoveDuplicates for LDEV-442

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-442
https://luceeserver.atlassian.net/browse/LDEV-3003


cfmitrah:
Skipped the second test in LDEV-3167(https://luceeserver.atlassian.net/browse/LDEV-3167), Because the error messages don't include the list of the available functions when the member function count reached 51. 

https://github.com/lucee/Lucee/blob/6.0/core/src/main/java/lucee/commons/lang/ExceptionUtil.java#L116

PR #1451, opened 2021-10-18
LDEV 3742 - CFContent delivers wrong Content-Type

Xcreen:
https://luceeserver.atlassian.net/browse/LDEV-3742

Set File-Mime-Type as Content-Type (only when type-attribute is not already set)


zspitzer:
@Xcreen the build failed due to some permissions problems, can you update your branch from upstream and it should pass?

PR #1454, opened 2021-10-20
Updated testcases with extended decimal values for LDEV-3746

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3746


michaeloffner:
manually adapted the changes because of merge conflicts

PR #1460, opened 2021-11-09
Added a testcase to check content-Type of cfcontent without type attribute for LDEV-3742

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3742


zspitzer:
@cfmitrah you can't use cfhttp in test cases, there's no http server when running them with script-runner

cfmitrah:
@zspitzer I disabled the testcase

zspitzer:
why disabled? just use `_internalRequest` ?

cfmitrah:
@zspitzer  _internalRequest() doesn't work with cfcontent like cfhttp. It doesn't return any result and abort the request at the location of _internalRequest().

I filed a bug for the issue https://luceeserver.atlassian.net/browse/LDEV-3761

PR #1466, opened 2021-11-18
LDEV-1925 Support UUID and Inet DB types in OtherCast

bdw429s:
LDEV-1925 Support UUID and Inet DB types in OtherCast

https://luceeserver.atlassian.net/browse/LDEV-1925

### DIFF
--- a/core/src/main/java/lucee/runtime/query/caster/OtherCast.java
+++ b/core/src/main/java/lucee/runtime/query/caster/OtherCast.java
@@ -22,6 +22,8 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.util.TimeZone;
+import java.util.UUID;
+import java.net.InetAddress;
 
 public class OtherCast implements Cast {
 
@@ -33,21 +35,38 @@ public class OtherCast implements Cast {
 
 	@Override
 	public Object toCFType(TimeZone tz, ResultSet rst, int columnIndex) throws SQLException {
-		if (type != Types.SMALLINT) return rst.getObject(columnIndex);
+		if (type != Types.SMALLINT) {
+			Object value = rst.getObject(columnIndex);
+			
+			// Drivers like Postgres like to return java.util.UUID instances instead of the string GUID 
+			if( value instanceof UUID ) {
+				return ((UUID)value).toString();
+			}
 
-		try {
-			return rst.getObject(columnIndex);
-		}
-		// workaround for MSSQL Driver, in some situation getObject throws a cast exception using getString
-		// avoids this
-		catch (SQLException e) {
+			// Drivers like Postgres have a custom type that returns java.net.InetAddress 
+			if( value instanceof InetAddress ) {
+				return ((InetAddress)value).toString();
+			}
+			
+			return value;
+				
+		} else {
+
 			try {
-				return rst.getString(columnIndex);
+				return rst.getObject(columnIndex);
 			}
-			catch (SQLException e2) {
-				throw e;
+			// workaround for MSSQL Driver, in some situation getObject throws a cast exception using getString
+			// avoids this
+			catch (SQLException e) {
+				try {
+					return rst.getString(columnIndex);
+				}
+				catch (SQLException e2) {
+					throw e;
+				}
 			}
 		}
+
 	}
 
 }


bdw429s:
@zspitzer I didn't actually change anythig in the try/catch below.  Github's diff is just confused since that code got indented to fit inside an "else" statement so it thinks it was changed.

core/src/main/java/lucee/runtime/query/caster/OtherCast.java
```diff
@@ -33,21 +35,38 @@ public OtherCast(int type) {
...
 	@Override
 	public Object toCFType(TimeZone tz, ResultSet rst, int columnIndex) throws SQLException {
-		if (type != Types.SMALLINT) return rst.getObject(columnIndex);
+		if (type != Types.SMALLINT) {
+			Object value = rst.getObject(columnIndex);
+			
+			// Drivers like Postgres like to return java.util.UUID instances instead of the string GUID 
+			if( value instanceof UUID ) {
+				return ((UUID)value).toString();
+			}
```
redtopia, line 44: The original issue indicated the problem was with UUID, but we ended up using GUID, which is also problematic. Can we do the same thing with GUID?

bdw429s: @redtopia Oh yeah, I meant to ask about that.  There is no GUID type in Postres that I can see-- at least not in their unit tests that go through and test all their data types.  THere is also no such java.util.GUID class, ONLY java.util.UUID.  Can you be more specific about what exact object you're getting back in your result sets?

redtopia: @bdw429s I think you are correct actually. I wasn't sure... my team concurs.

isapir:
These changes will break existing code.  Especially with the InetAddress I know that there are some places that expect the Postgres query to return an InetAddress object rather than a String.

These are custom features and are not part of any "standard" IMO.  If you want behavior like in other databases then use `text` as the type instead of Inet or UUID, or cast it to text in your queries.

bdw429s:
> there are some places that expect the Postgres query to return an InetAddress object rather than a String.

I've never seen anyone use the InetAddress, but the only time I've seen anyone use UUID, the only thing they did with the object was call `.toString()` on it, which would still work the same with a String.  

> These are custom features and are not part of any "standard" IMO.

The JDBC "standard" allows drivers to return whatever the heck they want.  The question is what is useful for CFML developers to use.  Lucee already had TONS of special exceptions for Oracle and its crazy return types.  I don't care if we want to remove the InetAddress if there are people actually calling methods (other than `.toString()`) on that object, but I don't think you'll find a person alive not wanting their UUID to be a string.  And Lucee already has a very strong precedent set for turning raw Java class instances from JDBC drivers into something useful here so that's nothing new.

redtopia:
I agree with @bdw429s on this, particularly for the UUID class. The precident with CFML is that UUID/GUIDs come back as strings. When we migrated from MSSQL to Postgres, there was no way we were going to go through the whole codebase to add a .toString() call whenever we wanted to access an ID field. So we ended up having to modify the postgres jdbc driver to make the .toString() call so that our code would just work. The alternative would have been to visit literally thousands of instances where we access a UUID from a query and then making sure we convert it before using it. That would have been a heavy lift, one that I would have not wanted to do because in my mind CFML should do it.

As far as the InetAddress goes, I can see it both ways... if you get an inet address back from a postgres db as a java object, I suppose you could then make java calls using that object. But isn't the point of using CFML to obfuscate that lower level java stuff? If your preference is to use java, couldn't you take the inet address string and convert it back to a java object in your CFML code? I think that is reasonable.

isapir:
I'm less concerned with the UUID type, but I am concerned about:

1. The slippery slope of adding ad-hoc patches to the Lucee code instead of returning what the JDBC driver returns
2. Converting an InteAddress to String just to convert it back to InetAddress - it's very inefficient and slow process

A simpler solution than hacking the PGJDBC driver would have been to use a `text` column type, or add a cast, or cast using the `::text` operator in your query.

If we hack the results from the driver then at the very least I would want an option that enables/disables that.

bdw429s:
> The slippery slope of adding ad-hoc patches to the Lucee code instead of returning what the JDBC driver returns

You're only calling it "ad hoc" because you don't like it.  The truth is Lucee has _always_ had an entire system of casting in place just for the case of massaging JDBC driver output to CFML types, including custom types for Oracle and Teradata.  This is part of the Lucee source code right.  CFML abstracts the JDBC layer away from the user and returns to them CFML data types where it makes sense, and it 100% makes sense for GUIDs.  

> Converting an InteAddress to String just to convert it back to InetAddress - it's very inefficient and slow process

First of all, prove it.  Java can create 10's of thousands of class instances with nearly no overhead.  Furthermore, your suggestion is that Lucee should return the java object and them we force the user to call `col.toString()` anyway, which uses java reflection to call native methods on java objects and I can totally prove to you that is much much slower than Lucee doing it guaranteed.  If you want to make this a debate of performance, then Lucee needs to convert it to a string, not the CFML code.

> A simpler solution than hacking the PGJDBC driver would have been to use a text column type

This is a strawman.  Lucee doesn't get to tell anyone what datatypes they use in their DB and it's not Lucee's place to do so.  Postgres supports a UUID type which is _more performant_ then a varcher because it stores it in binary as hex, not unicode chars.  It's Lucee's job to make this work seamlessly, not to tell people to re-design their databases.

> If we hack the results from the driver 

Again, it's not "hacking".  Lucee has an entire system of casting built in place to massage the raw Java objects that come out of JDBC drivers into something CFML can use natively.  

> I would want an option that enables/disables that.

I'm not against this idea, but if we take InetAddress of the table, I don't think you'll find a single person not converting their UUID to a string.  Heck, Adobe ColdFusion even does this, even for JSON types, so technically this is a compat issue.  Though it is worth noting that Adobe shows the UUID or PGObject in the query object dump, but converts to a string automatically when you go to use either of those objects.  I'm not sure where the equiv logic for that would be in Lucee, but that would be another option to do it like Adobe does.

redtopia:
We didn't use a text field instead of UUIDs because of performance, @bdw429s comment about allowing engineers to choose whatever data type makes sense for the app is exactly on point. We have tables with 10s of millions of rows and it was important to our team that we use UUIDs to ensure that our queries are as performant as possible. We did not spend the time to do tests because it seemed obvious (to us anyway) that UUIDs are going to be faster. We could have been wrong and maybe it would have been sixes, but I again agree with @bdw429s that limitations in Lucee shouldn't be the deciding factor.

If this enhancement is toggleable, I would suggest that the default behavior is to do the .toString() operation and then for any app that is dependent on the old behavior, they could add 1 line of code that will preserve their functionality.

zspitzer:
Returning UUID as string is a no brainer that's what CFML does

my question is, is the casting happening on accessing the field like with nulls?

if so then query.uuid_pk_col.getClass() provides access to the java object?

bdw429s:
> my question is, is the casting happening on accessing the field like with nulls?

I'm not clear whether you're asking about when Adobe CF turns the UUID class into a string or if you're asking about where the logic affected by this pull runs.  Since the answers are different, I'll address both possible questions.  Actually, I'm going to answer a third question as well because it covers all the bases.  

There are three specific points in time where data can be changed
* As the raw data is being read from the JDBC's result object and the CFML QueryImpl is being populated (this is part of the `fillResult()` method in `QueryImpl`
* As CFML code accesses the value in a given cell from a query (this is part of the `getValue()` method in `QueryUtil()`)
* As the value is used. i.e. output or compared to another value.  In the same manner that the `now()` function actually returns a `DateTimeImpl` instance, yet you can still use that as a string or compare it to other strings because Lucee casts it on the fly at the point of use.  (I THINK this happens in the many `toXXX()` methods in the `Caster` class.

So, with that said, here are the answers.  Null values in queries are covered to empty strings in bullet 2, but that doesn't really affect this conversation.  The code in this pull casts the UUID to a string as part of bullet 1, when the CFML queryImpl is being populated from the JDBC result.  HOWEVER, that said-- Adobe ColdFusion actually appears to convert the UUIDs to strings in bullet 3 at the point of use as part of the auto-casting system.

Now here's the interesting wrinkle which pinpoints _exactly_ where lucee differs from Adobe CF.  Check out this trycf example and run it both on Lucee and ACF 2021

https://trycf.com/gist/1a45c98339885ed05029596d7930f3c0/lucee5?theme=monokai
https://trycf.com/gist/40f1b6bf47beda60183e8f14c6e01699/acf2021?theme=monokai

You'll see that both Adobe CF and Lucee will _output_ a UUID instance as a string, but ONLY ACF will auto-cast on comparison to another string.  Lucee just blows up there.  

I think it would be fair to consider moving what I've done on this pull to be in bullet 3 instead of bullet 1.  Plus this is actually not even related to queries at all-- it's just a general purpose behavior of the language and how it handles UUIDs.  It's the same sort of effective outcome, BUT it leaves the raw Java instance of UUID in the result set until the point of use, which would mean people like Igal don't loose the ability to still call direct methods on that object if they wish.

bdw429s:
@michaeloffner Did you see the additional details and trycf links I added in my last comment?  I was actually suggesting we go a slightly different direction which was more in line with what Adobe CF does.  It would be a very similar result, but would leave the UUID instance in the query object and just auto-cast it upon comparisons.  In other words, I was suggesting we move the casting into the Caster instead of when building the QueryImpl instance (which I think would make Igal more happy as well)

bdw429s:
@zspitzer @michaeloffner Anyone?  What the heck is going on here?  We have a pull merged mid-discussion and then it just goes silent!

PR #1478, opened 2021-11-25
fix LDEV3496

softwareCobbler:
https://luceeserver.atlassian.net/browse/LDEV-3496

tldr:
One too many PageContext objects are (sometimes) pushed onto the stack. This fix updates a guard condition around an instance of `adapter.loadArg(0)` (which serves to push PageContext objects) to guard against situations where the pushed object won't be used.

Looks like the emitted bytecode for safenav constructs will occasionally place an unused  `PageContext` onto the stack, triggering the verify error when the JVM tries to unify frames from predecessor flows. Stepping through the codegen routines (VariableImpl.java) we can mark where where PageContexts get pushed; note that in the decompilation (via FernFlower) there are only 3 uses of `var1 : PageContext`, but there are definitely 4 pushes (we look at bytecode next):

```
// code for `isNull(foo?.bar)`
// 3 uses of var1, but stack gets 4 aload_1 
public final Object call(PageContext var1) throws Throwable {
    if (true) {
        IsNull.call(
            var1, // VariableImpl::_writeOutFirstBif:368
            CallerUtil.get(
                var1, // VariableImpl::_writeOutCallerUtil:292
                // VariableImpl::_writeOutFirstDataMember:653 ---> gets hit but doesn't show up in decompiled
                var1.us(), // VariableImpl::_writeOutFirstDataMember:666
                new Key[]{
                    this.keys[0],
                    this.keys[1]
                },
                (Object)null
            )
        );
    }

    return null;
}
```

Another hint that this is where to look is trying to use the "Javassist" analyzer to print out this method's frames (it throws an `InvocationTargetException: Operand stacks could not be merged, they are different sizes!`)

Since `isNull(foo?.bar)` is small we can trace the stack changes by hand (this is output from javap, but annotated with "current stack plus how this line will change it":
```
public final java.lang.Object call(lucee.runtime.PageContext) throws java.lang.Throwable;
descriptor: (Llucee/runtime/PageContext;)Ljava/lang/Object;
flags: ACC_PUBLIC, ACC_FINAL
Code:
stack=9, locals=2, args_size=2
0: iconst_1
[] + 1
1: ifeq          41
[1] - 1
4: aload_1           // push PageContext
[] + PC
5: aload_1           // push PageContext
[PC] + PC
6: aload_1           // push PageContext
[PC,PC] + PC          
7: aload_1           // push PageContext
[PC,PC,PC] + PC
8: invokevirtual #51                 // Method lucee/runtime/PageContext.us:()Llucee/runtime/type/scope/Undefined;

// ... snip ...

[PC,PC,PC,SCOPE,OBJ[2]] +null
34: invokestatic  #63                 // Method lucee/runtime/util/CallerUtil.get:(Llucee/runtime/PageContext;Ljava/lang/Object;[Llucee/runtime/type/Collection$Key;Ljava/lang/Object;)Ljava/lang/Object;
[PC,PC,PC,SCOPE,OBJ[2],null] !pop4, (push return)
37: invokestatic  #68                 // Method lucee/runtime/functions/other/IsNull.call:(Llucee/runtime/PageContext;Ljava/lang/Object;)Z
[PC,PC,Obj] !pop2 (push bool)
40: pop
[PC,bool] -bool
41: aconst_null (jump target from (1))
([PC]) | ([]) <---- stack frames do not unify -- from jump origin (line 1) it is [], from line 40 it is [PageContext]
42: areturn
[PC,null] | [null]
```

The issue appears to be line 41; the jvm wants either [PageContext] or [] as a stack, but it's getting both as possibilities.

It at least gets the the LDEV3496 test passing, I am interested in seeing if it passes the "real world" test!


softwareCobbler:
I think having merged in the recent testfixes it will pass the test suite.

michaeloffner:
@softwareCobbler great catch, thanks for your contribution!

PR #1480, opened 2021-11-28
fix LDEV-3333, LDEV-3731

softwareCobbler:
https://luceeserver.atlassian.net/browse/LDEV-3333
https://luceeserver.atlassian.net/browse/LDEV-3731

There are a few places where objects are placed into maps or sets in order to determine if they have been visited during some possibly-recursive algorithm. In these places, if the Object happens to extend `AbstractList` or similar, the default java hash implementation triggers a stack overflow, because hashing a list requires knowing the hashcode of all the list's elements. So if a list contains itself, it won't terminate.

This changeset updates instances of this pattern to use Maps or Sets keyed on `Integer`s derived from calls to `System.identityHashCode`.


michaeloffner:
@softwareCobbler great catch, love the implementation!

PR #1489, opened 2021-12-10
LDEV-3777 savecontent ignores whitespace management setting

Xcreen:
Use whitespace-management setting when trim-attribute is not set by the user.


michaeloffner:
@Xcreen thanks for your PR, great work!
i did modify it a little bit, hope you are fine with the change
https://github.com/lucee/Lucee/commit/604468d204a80a4ae9434c9831dcf344e3a6995c

PR #1496, opened 2021-12-22
LDEV-3369 Remove lowercase from PageSource-Mapkey, due of wrong path caching 

Xcreen:
https://luceeserver.atlassian.net/browse/LDEV-3369

Because of the lowercase of the key, wrong pagesources are returned for different paths.
This is not be a problem on case-insenstive filesystem, but leads to wrong paths on case-senstive filesystems.


michaeloffner:
This change would make the page pool case sensitive and cause multiple instances of the same page in the pool, but that is not expected. So let say you have this code
<cfinclude template="test.cfm">
<cfinclude template="Test.cfm">

If "test.cfm" is changed the second include would still show the old code, because the flush record would only flush one of the 2 records in the pool. This change would break the pool and blow it up. 

INDEPENDENT of the input the result should give the real name anyway, i assume this was not done because of performance reasons.

Thanks for your PR, but sorry we cannot accept this.

PR #1503, opened 2021-12-31
update Felix 6.0.5 for Java > 15

zspitzer:
update to https://luceeserver.atlassian.net/browse/LDEV-3526


JCCChris:
On: https://docs.lucee.org/guides/installing-lucee/download-and-install.html#supportedJavaVersions

It says "Java 16 is not currently supported due to breaking internal changes with the jvm..."  

Does this need to be updated?

zspitzer:
still not fully supported https://luceeserver.atlassian.net/browse/LDEV-3807

PR #1511, opened 2022-01-06
Added a fix: isValid(email) to return false rather than throw error for LDEV-3800, LDEV-2461, LDEV-3677

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3800
https://luceeserver.atlassian.net/browse/LDEV-2491
https://luceeserver.atlassian.net/browse/LDEV-3677

Note: this fix doesn't affect the current behavior of cfmail tag attributes(from, to, cc & bcc).


michaeloffner:
manually adapted the change, because of merge conflicts

PR #1512, opened 2022-01-08
LDEV-3810 add trace logging for CFHTTP calls (6.0)

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3810


michaeloffner:

PR #1515, opened 2022-01-13
Resolve LDEV-3736 & LDEV-3734

bdw429s:
This pull resolves both LDEV-3736 & LDEV-3734.  The logic is tightly intertwined, so the changes are combined.  There are full unit tests covering every code path for both tickets.

Note, I cleaned up some incorrectly-named methods internally since `^` is a bitwise operator, not an exponent.  I also added the `power()` function to complete the arithmetic operations and since this logic needed to apply to it as well so it all made sense to put it together.


bdw429s:
Fixes 
https://luceeserver.atlassian.net/browse/LDEV-3736
and
https://luceeserver.atlassian.net/browse/LDEV-3734

PR #1533, opened 2022-01-21
LDEV-1526 update hsqldb to 2.7.0 (QoQ)

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-1526


zspitzer:
only updating to 2.7.0 as 2.7.1 on maven doesn't have OSGI meta data published for jdk8

PR #1545, opened 2022-01-28
Added a testcase to QoQ with [] regex charsets in LIKE operator for LDEV-3826

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3826


bdw429s:
@cfmitrah This pull can be closed.  It is superseded by https://github.com/lucee/Lucee/pull/1546 which includes your test case plus many more added to it along with the fix.

PR #1546, opened 2022-01-29
LDEV-3826 QoQ LIKE support [char sets]

bdw429s:
Make Lucee QoQ `LIKE` match Adobe CF and MS SQL in supporting bracket char sets


bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-3826

PR #1547, opened 2022-01-29
LDEV-3851 build: test latest core with stable loader

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3851


zspitzer:
https://github.com/lucee/Lucee/pull/1548

PR #1558, opened 2022-02-03
LDEV-3864 add LUCEE_OSGI_FATAL env var

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3864

### DIFF
--- a/core/src/main/java/lucee/commons/lang/ClassUtil.java
+++ b/core/src/main/java/lucee/commons/lang/ClassUtil.java
@@ -21,6 +21,7 @@ package lucee.commons.lang;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PrintStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


core/src/main/java/lucee/commons/lang/ClassUtil.java
```diff
@@ -161,6 +164,15 @@ public static Class loadClassByBundle(String className, String name, Version ver
 			else ce = new ClassException(
 					"In the OSGi Bundle with the name [" + name + "] and the version [" + version + "] was no class with name [" + className + "] found." + appendix);
 
+			if (SystemUtil.getSystemPropOrEnvVar("lucee.osgi.fatal", "").equalsIgnoreCase("true")) {
+				String msg = "Lucee OSGI error. As LUCEE_OSGI_FATAL=true, shutting down.";
+				CONSOLE_ERR.println(e);
+				CONSOLE_ERR.println(msg);
+				//logDeploy.log(Log.LEVEL_ERROR, "osgi", msg);
+				//shutdownFelix();
+				System.exit(1);
```
michaeloffner, line 173: having a System.exit in the code for me is a no go, because at runtime this stops the servlet engine, so this should never be done nowhere.

michaeloffner:
@zspitzer as written above we cannot do that PR in this form, if yo still wanna do, plese reopen the PR with the necessary changes

PR #1577, opened 2022-02-23
Added a fix debugging cache Type to show when Database Activity enabled for LDEV-1673

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1673


michaeloffner:
this commit turns the flag "debugQueryUsage" useless and log it always. Solution is not to show that info at all in that case.

PR #1583, opened 2022-03-03
Add failing test for getMetadata on UDFs

elpete:
Lucee does not respect a `@description` in a docblock but does as a function annotation.

I tried to find where I would adjust this in code but no luck.  Hopefully the failing test case helps.

### DIFF
--- a/test/functions/GetMetaData.cfc
+++ b/test/functions/GetMetaData.cfc
@@ -62,7 +62,7 @@ component extends="org.lucee.cfml.test.LuceeTestCase" {
 				assertEquals("soft",meta.type);
 			});
 
-			it( "Checking UDFs", () => {
+			it( title = "Checking UDFs", body = () => {
 				var metaA = getMetadata( exampleFunctionWithDescriptionInAnnotation );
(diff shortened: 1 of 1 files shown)


bdw429s:
@elpete Is there a ticket for this?

elpete:
Probably not. I was just tinkering with it with some cbvalidation changes.

bdw429s:
Cool, I'll create one.  I think I MAY have found the logic in Lucee that deals with this.  "description" is some sort of special annotation the parser looks for and it has some special handling.  I couldn't confirm it however without creating a custom compiled version of Lucee with debugging in place and I didn't care about it that much, lol.

bdw429s:
Ticket: https://luceeserver.atlassian.net/browse/LDEV-3899

zspitzer:
@elpete can you rebase (not via the web ui, it doesn't trigger a build) this PR, once it's run thru a build, then add `labels="metadata"` and `skip="true"` so I can merge it?

elpete:
@zspitzer Done.

PR #1590, opened 2022-03-11
LDEV-3264 add test cases for onSessionEnd

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3264


zspitzer:
**do not merge**, there is a filter on test cases to run just this test case

PR #1606, opened 2022-03-23
Added a adminMode key to GetApplicationSettings() and server.lucee scope for LDEV-3916

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3916


zspitzer:
@cfmitrah `server.lucee.adminmode` is meh... lets use `server.lucee.singleContext=true/false`

PR #1609, opened 2022-03-29
Added a fix to QueryAddRow to handle struct of array correctly for LDEV-3933 (lucee 5.3)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3933


bdw429s:
@cfmitrah Thanks for taking a stab at this.  To be honest, I'm not sure if it makes sense to have a different behavior between `querynew()` and `queryAddRow()`.  I'm tempted to say they should both work the same.  I'd be curious to hear what Micha says on this.

PR #1617, opened 2022-04-13
Improved the message for output the component with debug enabled

cfmitrah:


adamcameron:
Note there's a grammar error in the updated message:
```
To see details for this component you need to enable debug output and create debugging template
```

Should read:
```
To see details for this component you need to enable debug output and create a debugging template
```

PR #1619, opened 2022-04-13
LDEV-3659 test cases for mixed orm and query 

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3659

passes with 3.5.5.84-SNAPSHOT
hangs with 5.4.29.13-BETA


zspitzer:
updated to test both h2 and mssql datasources( timeout stops hang )

manifest in this PR is using Hibernate 5.4 to show the problem and only runs this test case (for now)

PR #1628, opened 2022-05-03
Fix string typo

KamasamaK:
Fix typo in Lucee Admin for Regex Dialect description.


isapir:
Thanks @KamasamaK!  And thank you for your work on the VS Code extension!

PR #1642, opened 2022-05-16
Added a fix implement the object.toString() member function and test to CreateODBCTime().toString() for LDEV-3993

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3993


michaeloffner:
@cfmitrah that approach is to broad and will break other stuff for sure. that needs a more specialised approach.

PR #1644, opened 2022-05-17
Added a testcase to cfschedule with userAgent attribute for LDEV-2999

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2999


zspitzer:
@cfmitrah thing is, this isn't going to work as there's no http server when the unit tests are run....

PR #1651, opened 2022-05-27
Added a test to cflog without file attribute for LDEV-3978

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3978

### DIFF
--- a/test/tickets/LDEV3978.cfc
+++ b/test/tickets/LDEV3978.cfc
@@ -2,30 +2,32 @@ component extends="org.lucee.cfml.test.LuceeTestCase" labels="logs" {
 
 	function beforeAll() {
 		variables.filePath = "#expandPath("{lucee-config}")#/logs";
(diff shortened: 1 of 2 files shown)


tomchiverton:
When I write tests that look for strings like `expect(findNoCase("test_application_without_file_second", appLog))` I always try and include a UUID so I know I'm not looking at old results in the log file.

I don't know if the Lucee test infrastructure removes logs before every run or similar that would negate this.

cfmitrah:
@tomchiverton 
> I don't know if the Lucee test infrastructure removes logs before every run or similar that would negate this.

 before starting the build process, lucee deletes the temp directory ( which has the lucee-server, webroot, etc ) created in the previous build.
https://github.com/lucee/Lucee/blob/e543a45283a93a0da177369af0eb158c2c6dd276/ant/build-core.xml#L486


I updated the testcase to use UUID in text

PR #1669, opened 2022-06-10
unhide the datetime functions and testcase for LDEV-3115

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3115


michaeloffner:
see the ticket comment for why this is rejected

PR #1677, opened 2022-06-20
LDEV-3863 additional empty param test cases

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3863


zspitzer:
anyone know if ACF throws an error when a named parameter in a query doesn't exist? Lucee seems to be casting it to null, which feels super wrong, at least here with QoQ with the first added test here

the second one throws in the native Lucee QoQ and then tries just to fall back to hsqldb, which throws an error

zspitzer:
closing this as we have raised a new ticket https://luceeserver.atlassian.net/browse/LDEV-4044

PR #1695, opened 2022-07-08
Added a fix to cftimer to support console type for LDEV-4066

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4066


zspitzer:
@cfmitrah wow that was quick! No need for the timer prefix, I think?

cfmitrah:
> @cfmitrah wow that was quick! No need for the timer prefix, I think?

@zspitzer  I've updated the PR to use the CFTimer prefix when there is no label

PR #1708, opened 2022-07-19
Added a fix to catch the debugging templates error in admin and always get the duplicate result of debug data for LDEV-4087

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4087


cfmitrah:
This issue was fixed in Lucee versions 5.4, 6.0, and 6.1. Therefore, I am closing this PR in 5.3.

PR #1717, opened 2022-07-22
LDEV-3877 only check for empty query param when actually used

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3877

replaces https://github.com/lucee/Lucee/pull/1573

### DIFF
--- a/core/src/main/java/lucee/runtime/tag/util/QueryParamConverter.java
+++ b/core/src/main/java/lucee/runtime/tag/util/QueryParamConverter.java
@@ -231,14 +231,14 @@ public class QueryParamConverter {
 		SQLItems<NamedSQLItem> item;
 		while (it.hasNext()) {
 			item = it.next();
-			if (item.isEmpty()) {
-				throw new ApplicationException("param [" + name + "] may not be empty");
-			}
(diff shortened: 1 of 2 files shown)


core/src/main/java/lucee/runtime/tag/util/QueryParamConverter.java
```diff
@@ -231,14 +231,14 @@ private static SQLItems<SQLItem> get(String name, List<SQLItems<NamedSQLItem>> i
...
 			item = it.next();
-			if (item.isEmpty()) {
-				throw new ApplicationException("param [" + name + "] may not be empty");
-			}
-			if (item.get(0).name.equalsIgnoreCase(name)) {
+			if (item.size() > 0 && item.get(0).name.equalsIgnoreCase(name)) {
+				if (item.isEmpty()) {
+					throw new ApplicationException("queryparam [" + name + "] may not be empty");
+				}
 				return item.convertToSQLItems();
```
michaeloffner, line 238: old version first check if empty or not, if not empty it get the value of the first element.
new version check if empty after it already has checked the size>0, what is the same as not empty.
That makes no sense.

michaeloffner:
PR makes no sense in this form, please reopen issue when improved

PR #1718, opened 2022-07-22
Added support for passing LUCEE_BUILD_ENV to ANT

dswitzer:
I've added support for passing the LUCEE_BUILD_ENV environmental variable to ANT, so you can specify it on the command line when building Lucee.


dswitzer:
This update just adds support for passing in the `LUCEE_BUILD_ENV` value in the ant command line, so that you can do:

ant.bat -DLUCEE_BUILD_ENV="c:/work/lucee_build_env.json"

michaeloffner:
this is already in the code

PR #1719, opened 2022-07-23
LDEV-3970  default to returning json rather than wddx

zspitzer:
Lucee doesn't even start with this change ATM
https://luceeserver.atlassian.net/browse/LDEV-3970


michaeloffner:
i have used this PR as template but added support for the application.cfc setting "this.returnFormat", more details in the ticket

PR #1720, opened 2022-07-23
LDEV-4081 FileWrite exception is missing the filepath (again)

zspitzer:
this causes side effects, see task

But doesn't fail with a PR as s3 not available

https://luceeserver.atlassian.net/browse/LDEV-4081


michaeloffner:
@cfmitrah can you please resolve the conflict

michaeloffner:
manually adapted the change

PR #1732, opened 2022-07-29
updated a fileWrite fix to works only for fileResource for LDEV-4081

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4081


michaeloffner:
this is already applied in more or less the same way like this
`// validate parent only works when you have access to the parent, that is not necessary a given for
				// all FS
				if ("file".equalsIgnoreCase(res.getResourceProvider().getScheme())) {
					Resource parent = res.getParentResource();
					if (parent != null && parent.canRead() && !parent.exists()) throw new ApplicationException("parent directory for [" + res + "] doesn't exist");
				}`

PR #1735, opened 2022-07-29
LDEV-4117 Add ldap service and tests to build

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4117

problem i'm getting this error

`ERROR Service [ ldap ] threw [ simple bind failed: localhost:10389 ]`

think I got the creds right? last time i used CFLDAP was the late 90s!

https://github.com/intoolswetrust/ldap-server#default-ldif

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,11 @@ jobs:
           - 3993:3993 #IMAPS
           - 3995:3995 #POP3S
           - 8080:8080 #API
+      ldap:
+        image: kwart/ldap-server
+        ports:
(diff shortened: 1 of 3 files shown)


zspitzer:
the build is green due to non available services being skipped by default

https://github.com/lucee/Lucee/runs/7581622275?check_suite_focus=true#step:20:429

zspitzer:
I have commented out LDAP_PASSWORD in `.github\workflows\main.yml` which disables the LDAP service

I will merge this in, so anyone can fork 6.0, uncomment that line and play around if they want to help out

PR #1748, opened 2022-08-04
Added a fix to reMatch/reMatchNoCase regression with empty string for LDEV-3703 (lucee 5.3.9)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3703


zspitzer:
see https://github.com/lucee/Lucee/pull/1621 too

PR #1767, opened 2022-08-17
LDEV-4154 hash (quick) with numIterations > 1  don't log, just throw

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4154


bdw429s:
@zspitzer See my comment on the ticket, I don't agree with this change.

bdw429s:
@michaeloffner @zspitzer No response to my concerns at all?  We've thrown the baby out with the bathwater here.  Why not just implement the iterations so the function does what it says it does?

PR #1799, opened 2022-09-15
Added a fix to implement the support for method argument in InternalRequest() LDEV-4199

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4199

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
+++ b/core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
@@ -67,7 +67,7 @@ public class InternalRequest implements Function {
 			boolean addToken, boolean throwonerror) throws PageException {
 
 		method = method.toUpperCase().trim();
-		if (methods.indexOf(method) < 0) throw new FunctionException(pc, "_InternalRequest", 2, "method", "invalid method type [" + method + "], valid types are [" + methods.toString() + "]");
+		if (methods.indexOf(method) < 0) throw new FunctionException(pc, "_InternalRequest", 2, "method", "invalid method type [" + method + "], valid types are [" + ListUtil.arrayToList(methods.toArray(new String[0]), ",") + "]");
 		
 		Struct urls = toStruct(oUrls);
 		Struct forms = toStruct(oForms);
@@ -89,7 +89,7 @@ public class InternalRequest implements Function {
 		String ext = ResourceUtil.getExtension(template, null);
 		// template
 		if (StringUtil.isEmpty(ext)) {
-			throw new FunctionException(pc, "_InternalRequest", 1, "template", "template path is invalid");
+			throw new FunctionException(pc, "InternalRequest", 1, "template", "template path is invalid");
 		}
 
 		// dialect
--- a/core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
+++ b/core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
@@ -5,7 +5,9 @@ import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
```diff
@@ -58,8 +60,15 @@ public class InternalRequest implements Function {
...
 	private static final Key CONTENT_LENGTH = KeyImpl.getInstance("content-length");
 
+	private static final List<String> methods = Arrays.asList(new String[] { "GET", "POST", "HEAD", "PUT", "DELETE", "OPTIONS", "TRACE", "PATCH" });
+
+
 	public static Struct call(final PageContext pc, String template, String method, Object oUrls, Object oForms, Struct cookies, Struct headers, Object body, String strCharset,
 			boolean addToken, boolean throwonerror) throws PageException {
+
+		method = method.toUpperCase().trim();
+		if (methods.indexOf(method) < 0) throw new FunctionException(pc, "_InternalRequest", 2, "method", "invalid method type [" + method + "], valid types are [" + methods.toString() + "]");
```
michaeloffner: That .toString() will not work as expected, you need to do the following instead ListUtil.arrayToList(methods,", ")

core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
```diff
@@ -78,9 +87,9 @@ public static Struct call(final PageContext pc, String template, String method,
 		Charset reqCharset = StringUtil.isEmpty(strCharset) ? pc.getWebCharset() : CharsetUtil.toCharset(strCharset);
 
 		String ext = ResourceUtil.getExtension(template, null);
-		// welcome files
+		// template
 		if (StringUtil.isEmpty(ext)) {
-			throw new FunctionException(pc, "Invoke", 1, "url", "welcome file listing not supported, please define the template name.");
+			throw new FunctionException(pc, "_InternalRequest", 1, "template", "template path is invalid");
 		}
```
michaeloffner, line 93: For Lucee 6 we did add "InternalRequest" asn an alias for "_InternalRequest" so when we do the same for 6 we need to change the name there

core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
cfmitrah: @michaeloffner thanks for reviewing this PR and input on the changes
 I've modified it to use ListUtil.arrayToList() https://github.com/lucee/Lucee/pull/1799/commits/b30c0de34351b6babfbc0ad4e3e014c88a33205f

can you please review my changes?

core/src/main/java/lucee/runtime/functions/system/InternalRequest.java
cfmitrah: @michaeloffner  thanks for reviewing this PR and input on the changes
I've changed the name to InternalRequest https://github.com/lucee/Lucee/commit/b30c0de34351b6babfbc0ad4e3e014c88a33205f

core/src/main/java/lucee/runtime/tag/Http.java
```diff
@@ -635,17 +635,12 @@ public void setDomain(String domain) {
 	public void setMethod(String method) throws ApplicationException {
 		method = method.toUpperCase().trim();
 		short idx = (short) methods.indexOf(method);
-		if (idx < 0) throw new ApplicationException("invalid method type [" + method + "], valid types are [" + methods.toString() + "]");
+		if (idx < 0) throw new ApplicationException("invalid method type [" + method + "], valid types are [" + ListUtil.arrayToList(methods.toArray(new String[0]), ",") + "]");
 		this.method = idx;
```
michaeloffner, line 639: ListUtil should have a method in my opinion that accepts as inputs a list

PR #1804, opened 2022-09-19
Added a fix to set to connectionLimit default (-1) in AdminUI for LDEV-1445 (lucee 5.3)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1445
https://luceeserver.atlassian.net/browse/LDEV-4209


core/src/main/java/lucee/runtime/config/XMLConfigAdmin.java
```diff
@@ -1665,7 +1665,7 @@ else if (access >= SecurityManager.VALUE_1 && access <= SecurityManager.VALUE_10
 		if (!StringUtil.isEmpty(timezone)) el.setAttribute("timezone", timezone);
 		el.setAttribute("database", database);
 		if (port > -1) el.setAttribute("port", Caster.toString(port));
-		if (connectionLimit > -1) el.setAttribute("connectionLimit", Caster.toString(connectionLimit));
+		el.setAttribute("connectionLimit", Caster.toString(connectionLimit));
```
michaeloffner, line 1668: in code i revisit with this, i do `else el.removeAttribute("connectionLimit")`, but i think it is fine,because with Lucee 6 this completely change anyway.

PR #1805, opened 2022-09-19
Added a fix to always include a charset key in cfhttp result for LDEV-3678

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3678


michaeloffner:
@cfmitrah is that still needed? the ticket is already closed.

zspitzer:
this is also against 5.3 :)

michaeloffner:
this is already implemented and ticket is already closed

PR #1806, opened 2022-09-19
QoQ shouldn't fall back to HSQLDB after queryparam error

bdw429s:
This pull adds a new exception type `IllegalQoQException` which indicates a QoQ is illegal and needs to fail right away without attempting to fail back to HSQLDB which will obfuscate the original error.  I have added test cases for every place QoQ currently fails and we know there is no reason to try again with HSQLDB.

### DIFF
--- a/core/src/main/java/lucee/runtime/exp/IllegalQoQException.java
+++ b/core/src/main/java/lucee/runtime/exp/IllegalQoQException.java
@@ -1,0 +1,44 @@
+/**
+ *
+ * Copyright (c) 2014, the Railo Company Ltd. All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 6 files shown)


bdw429s:
@zspitzer @micstriit https://luceeserver.atlassian.net/browse/LDEV-3878

bdw429s:
The following QoQ errors which used to fall back to HSQLDB will now throw the original, and much more useful exception, directly from the native QoQ implementation.

* Divide by zero - `1/0`
* Mod zero - `1%0`
* Too many count columns - `count( col1, col2 )`
* Too many count distinct columns - `count( DISTINCT col1, col2 )` 
* Cast with missing type - `cast( col )`
* Union with mismatched column counts in the selects
* Union with invalid order by 
* Order by with invalid ordinal position - `ORDER BY 9001`
* Positional param value with wrong type
* Named param value with wrong type

bdw429s:
@zspitzer Once this first pull is merged and the commits get rebased into the other pull, it will pass as well since it uses some of this new code (the ability to prevent the HSQLDB fallback)

core/src/main/java/lucee/runtime/exp/IllegalQoQException.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
```
michaeloffner, line 18: new classes not need the RAilo copyright...

bdw429s: @michaeloffner Got it!  Thanks for letting me know, I just copied it all from another class.

PR #1809, opened 2022-09-21
LDEV-3652 SerializeJSON ignores remotingFetch and SerializeJSON does not call custom getter function for component properties

mpark4656:
SerializeJSON function should not ignore remoting fetch. SerializeJSON function should invoke the getter function of a component's property if one is defined for that property.

https://luceeserver.atlassian.net/browse/LDEV-3652


mpark4656:
Application.cfc must have this.triggerDataMember set to true for the SerializeJSON function to access the getter functions. If triggerDataMember is not true, then SerializeJSON will work the way it did before this change.

core/src/main/java/lucee/runtime/functions/conversion/SerializeJSON.java
```diff
@@ -84,7 +84,7 @@ else if (!StringUtil.isEmpty(useSecureJSONPrefixOrCharset)) {
 
 	private static String _call(PageContext pc, Object var, Object queryFormat, Charset charset, boolean useSecureJSONPrefix) throws PageException {
 		try {
-			JSONConverter json = new JSONConverter(true, charset, JSONDateFormat.PATTERN_CF);
+			JSONConverter json = new JSONConverter(false, charset, JSONDateFormat.PATTERN_CF);
```
zspitzer, line 87: this jumps out to be as being wrong or at least incomplete?

zspitzer (changes requested):

zspitzer:
@mpark4656 i raised a question about the wholesale changing of the true to false in the call to JSONConverter?

the test case is a good start, but we need tests to cover all the different behaviours, not just this specific use case

core/src/main/java/lucee/runtime/functions/conversion/SerializeJSON.java
mpark4656: Hello,

JSONConverter constructor's first argument is a boolean flag for ignoreRemotingFetch. It basically determines if the converter should ignore the remotingFetch attribute on a property. In JSONConverter.java, this flag is referenced when the converter is converting a component, which is where the remotingFetch attribute would be used on a property. Since the goal is to make the SerializeJSON function not ignore that property attribute, I am passing 'false' for the first argument there.

I will add more test cases. SerializeJSON can be used on arrays, structs, nested structs, and etc., so I will start there.

Thank you.

mpark4656: @zspitzer I committed & pushed additional test cases to cover different behaviors of the SerializeJSON function.

Thank you.

michaeloffner:
@mpark4656 @zspitzer where are we on this PR, @zspitzer can we merge it in your opinion?

zspitzer:
@michaeloffner looks good to me, @mpark4656 has done great job with the test coverage

michaeloffner:
i had to manually adapt this change, because merging this to 6.1 was nearly impossible, to much changed
https://github.com/lucee/Lucee/commit/ad99628b5a2a528742f4b55c28a49d2f4771a68c

PR #1816, opened 2022-09-26
Fix for LDEV-4044 (again)

bdw429s:
Re-sending this pull due to issues rebasing the original one


bdw429s:
This fixes the case where named params are used in a QoQ but no params are provided to the query.

https://luceeserver.atlassian.net/browse/LDEV-4044

bdw429s:
Re-based with 5.3 to try and fix tests

PR #1817, opened 2022-09-27
Added a fix to cftimeout tag to throws error only when the exception occurred for LDEV-4215

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4215


cfmitrah:
Solved by this commit https://github.com/lucee/Lucee/commit/278908dce227f7dc4e4d171bb3208f2a3c2b016a

PR #1832, opened 2022-10-03
Schedule task fails between threads

duranguenze:
When i add a schedule task it marks fail on line 226, because the value is lower than task length, the length for the for, its pushed to the heap and do not read again in the cicle, better use a function or validate in each cicle.


cfmitrah:
@duranguenze thanks for this Pull Request.

Can you please create a new ticket in the [Lucee bug tracker](https://luceeserver.atlassian.net/) for the issue you faced with the details? after that please add the Pull Request link on the ticket's comment. :slightly_smiling_face:

michaeloffner:
@duranguenze i see what you tried to achieve, your solution is a good start but only minimize the issue adn does not address the underlying issue.
I have rewritten the code, so it uses a thread safe collection https://github.com/lucee/Lucee/commit/f864d79a8b0d7a7ce29a2d0f8a1d0d8cce660afb

michaeloffner:
@duranguenze thanks for your input

PR #1864, opened 2022-11-01
LDEV-3947 http resource exists() stricter status code checks

zspitzer:
only return exists true when status code is between 200-299

https://luceeserver.atlassian.net/browse/LDEV-3947


michaeloffner:
manually adapted this changed.https://github.com/lucee/Lucee/commit/5c405e7a32b190dd2b460ec0f11b14ea92d25431

PR #1887, opened 2022-11-30
LDEV-4298 QoQ performance enhancements

bdw429s:
This pull should not change the outward behavior of QoQ at all.  It refactors all the looping does internally the process a native QoQ to instead use Java Streams.  If there are more than 50 records in the query object, a parallel stream will be used which uses more than one thread to process your query.  This brings a performance boost of over 200% which I've tested using up to 1 Million rows.  The more complex the query is (where clause operations, selects, grouping, etc) the better the performance gains are.  Lucee is now also now even more performant than Adobe CF's QoQ in nearly all cases.  

The threshold of parallelism can be controlled by the following new system property: `lucee.qoq.parallelism` 
Set the property to the number of records where you want parallel threads to kick in.  Setting it to `0` would make all QoQs parallel. setting it to a large number like `9999999` would effectively make no QoQ's parallel.  The default is 50, which is where I measured the performance gains outweighed the overhead of threading.

Note, Java Streams will not use multi-threaded sorting (used in `ORDER BY` processing) until the number of items is above ~8,000. This is a hard-coded limit baked into the JDK.

Java streams use the fork/join pool by default, which has as many threads as you have CPU cores on the machine.  Since QoQ is an inherently CPU-bound operation, this makes sense to use.  This also means that your performance improvements will be directly tied to your number of CPU cores because more cores means more threads in the pool.  

The `ORDER BY` logic has also been entirely rewritten.  The sort is no longer performed on each column in separate passes, which copies all data in the query for every pass.  There is now a new `QueryComparator` class which sorts all columns in the `ORDER BY` in a single pass, short circuiting where possible so if the first column is already sorted above or below, it doesn't even proceed to the next column.  Furthermore, the sort register logic has been removed in favor of an `int[]` stream which maps the original row to the new sorted location without the need to have any copies of the data nor a separate sort register class to track the mapping.  The actual data in the query is only re-sorted in memory once.  This has around a 300% increase in sort performance when several columns are being sorted on.  Many new tests have been added to ensure the behavior of sorting every possible type of query column behaves the same as Lucee 5.3.10.

Other change have been made to reduce async bottlenecks when multiple threads are adding rows to a query object.  The addRow() method was previously synchronized on both the query and column class, which led to unecessary blocking since every addRow() call doesn't necessarily touch the native array of data.  I have removed the heavy-handed syncronization from these methods, opting to ONLY syncronize the portion of the `growTo()` method IF AND ONLY IF the internal native array needs grown.  In all other cases, there is no locking.  Furthermore, I have replaced the `recordcount` property of the `QueryImpl` class and the `size` property of the `QueryColumnImpl` classes to be an `AtomicInteger` to ensure they don't get corrupted without the locking in a multi-threaded environment.  I have also modified `getRow()` to return the row specifically added, not just the last row in the query as the previous behavior was  not threadsafe.

The class that handles compilation of `LIKE` statements now employs a simple double-checked synchronized block to prevent parallel streams from compiling the same regex twice.  

Special treatment has also been added to the `ColumnExpression` class so it no longer holds cached references to the original query object when it is not safe to do so in multiple threads

It was necessary add new methods to the `QueryImpl` and `QueryColumnImpl` classes for some of this functionality.  Since the `Query` and `QueryColumn` interfaces are in the loader, which is not being changed for Lucee 6, I was forced to switch the typing in the `QoQ` class to use the concrete class.  Otherwise, I would have to constantly cast the query objects to `QueryImpl` to use these methods and I was concerned about the performance (not to mention the code readability).  When Lucee updates the loader again, we can add these new methods to the interfaces and go back and start using the Interface types.

I also noticed heavy string concatenation performance was VERY poor under load due to Lucee always trying to cast both operands to a number, and when that failed, THEN defaulting to string concat.  I remedied that to use `Decision.isNumber()` which is much faster to detect when we have strings or numbers.  This had a huge performance boost for any expressions concatenating strings in the select, group by, or where.

All previous QoQ tests should be passing.  I have added a new suite of tests that hits all affected QoQ functionality and that runs against 1 Million records to ensure the parallel threads kick in.  

Note: when you use a QoQ and do NOT use an `ORDER BY` clause, you cannot depend on the final rows to be in the same order as the original query due to the asynchronous nature of threading.  If you want a specific order, you'll need to use an `ORDER BY`.  The previous QoQ implementation tended to preserve the order of the original query object, but this was never a guarantee.


bdw429s:
https://luceeserver.atlassian.net/browse/LDEV-4298

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -411,39 +406,67 @@ private void executeSingleNonPartitioned(PageContext pc, Select select, Query so
...
 		}
 
-		// Loop over all rows in the source query
-		for (int row = 1; row <= source.getRecordcount(); row++) {
-			// Does this do anything??
-			sql.setPosition(0);
-			// If we can, optimize the max rows exit strategy
-			if (hasMaxrow && maxrows <= target.getRecordcount()) break;
-
-			// The where clause is a single Operation expression that returns try or false. Does
```
animex, line 428: true or false.

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -411,39 +406,67 @@ private void executeSingleNonPartitioned(PageContext pc, Select select, Query so
...
-			// this row match the where clause, if any?
-			if (where == null || Caster.toBooleanValue(executeExp(pc, sql, source, where, row))) {
-				target.addRow(1);
-				// If we have a match, add this row into the target query
+		IntStream stream = getStream( source );
+		if( where != null ) {
+			stream = stream
+				.filter( throwingFilter(  row-> {
+					// The where clause is a single Operation expression that returns try or false.
```
animex: true or false.

core/src/main/java/lucee/runtime/db/QoQ.java
```diff
@@ -481,33 +504,20 @@ private void executeSinglePartitioned(PageContext pc, Select select, Query sourc
...
-			// If where operation is matched (or doesn't exist) ....
-			if (where == null || Caster.toBooleanValue(executeExp(pc, sql, source, where, row))) {
-				// ... add this row to our partitioned data
-				queryPartitions.addRow(pc, source, row, false);
-			}
+		IntStream stream = getStream( source );
+		if( where != null ) {
+			stream = stream
+				.filter( throwingFilter(  row-> {
+					// The where clause is a single Operation expression that returns try or false.
```
animex: true or false.

test/tickets/LDEV4298.cfc
```diff
@@ -0,0 +1,263 @@
+component extends="org.lucee.cfml.test.LuceeTestCase" labels="qoq" {
+	variables.typeList = 'BIGINT,BINARY,BIT,BOOLEAN,CHAR,DATE,DECIMAL,DOUBLE,INTEGER,LONGVARBINARY,LONGVARCHAR,NCHAR,NUMERIC,NVARCHAR,REAL,SMALLINT,TIME,TIMESTAMP,TINYINT,VARBINARY,VARCHAR';
+
+	function beforeAll() {
+		employees = queryNew( 'name,age,email,department,isContract,yearsEmployed,sickDaysLeft,hireDate,isActive,empID,favoriteColor', 'varchar,integer,varchar,varchar,boolean,integer,integer,date,boolean,varchar,varchar' );
+		// 1 Million records
+		loop from=1 to=62500 index='local.i' {
+			employees.addRow( [
+				[ 'John Doe#i#',28,'John#i#@company.com','Acounting',false,2,4,createDate(2010,1,21),true,'sdf','red' ],
```
animex: 'Accounting'

core/src/main/java/lucee/runtime/sql/exp/ColumnExpression.java
```diff
@@ -125,19 +131,55 @@ public int getColumnIndex() {
...
+		Tells this column expression to not cache the column reference back to the original query
+	*/
+	public void setCacheColumn(boolean cacheColumn) {
+		this.cacheColumn = cacheColumn;
+	}
+
+	/**
+		Acquire the actual query column reference, taking caching into account
+		We cache the lookup of the column for basic selects because we run the same thing
+		over and over on the same query object.  But for partitioned selects, we have multuple query
```
animex: multiple query

core/src/main/java/lucee/runtime/type/comparator/QueryComparator.java
```diff
@@ -0,0 +1,145 @@
...
+import lucee.runtime.functions.other.Dump;
+import lucee.runtime.type.Collection.Key;
+import lucee.runtime.type.Query;
+import lucee.runtime.type.QueryColumn;
+import lucee.runtime.db.SQL;
+import lucee.runtime.type.Collection;
+import lucee.runtime.type.KeyImpl;
+
+/**
+ * Implementation of a Comparator that will sort multuple rows of a query all at the same time
```
animex: multiple rows

bdw429s:
@animex Thanks for the spelling fixes.

PR #1918, opened 2022-12-22
Added a testcase to check cfqueryparam with scale attribute for LDEV-4337

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4337


zspitzer:
why use internal request?

PR #1921, opened 2022-12-23
Update server.regional.cfm

andreasRu:
Add information that  "server date/time" is the JVM time and not related to the server-context timezone setting. 
See this thread: https://dev.lucee.org/t/server-admin-is-showing-a-contexts-date-time-why/11775/7


andreasRu:
Please excuse that I didn't create an LDEV ticket. That is such a minor change that I didn't want to bother anybody with this. If you want you can of course reject and silently close this.

zspitzer:
nice tweak!

no need for a ticket, but can you change that in the lang files instead, ie en.xml etc

andreasRu:
@zspitzer I've corrected and added it to all available language files.
Merry Christmas to you all!

PR #1927, opened 2023-01-06
LDEV-4358 fix resource leak opening db connections

darrell-rg:
When getDatasourceConnection() decides to create a new connection, it first calls _inc() to increment the connection count. If loadDatasourceConnection() throws a page error, _dec() is called. But if the thread is killed by a timeout _dec() never gets called and so we lose a connection slot. If enough connection slots are lost lucee will not be able to recover when the db comes back.

This changes getDatasourceConnection() so that it only calls _inc() after the connection is successfully opened. This is not a perfect solution since it could allow the ConnectionLimit to be temporarily exceeded.

Also DCStack.isValidEL was doing an incorrect conversion from miliseconds to seconds, that ended up using a timeout of zero.    DataSourceSupport.NETWORK_TIMEOUT_IN_SECONDS  is set at 10,   so a conversion is not needed.   

See https://dev.lucee.org/t/resource-leak-in-datasourceconnectionpool-5-3-10/11853

https://luceeserver.atlassian.net/browse/LDEV-4358

### DIFF
--- a/core/src/main/java/lucee/runtime/db/DCStack.java
+++ b/core/src/main/java/lucee/runtime/db/DCStack.java
@@ -214,11 +214,9 @@ class DCStack {
 
 	private Boolean isValidEL(Connection conn) {
 		try {
-			// value is in ms but method expect s
-			int ms = datasource.getNetworkTimeout();
-			int s = DEFAULT_TIMEOUT;
-			if (ms > 0) s = (int) Math.ceil(ms / 1000);
-
+			int s = datasource.getNetworkTimeout();
+			//if timeout is unreasonable, use default
+			if (s < 1 || s > 60) s = DEFAULT_TIMEOUT;	
 			return conn.isValid(s) ? Boolean.TRUE : Boolean.FALSE;
 		}
 		catch (Exception e) {
--- a/core/src/main/java/lucee/runtime/db/DatasourceConnectionPool.java
+++ b/core/src/main/java/lucee/runtime/db/DatasourceConnectionPool.java
@@ -110,18 +110,17 @@ public class DatasourceConnectionPool {
 						break;
 					}
 				}
-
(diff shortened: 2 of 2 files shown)


core/src/main/java/lucee/runtime/db/DCStack.java
```diff
@@ -214,11 +214,9 @@ private boolean isClosedEL(Connection conn) {
...
 		try {
-			// value is in ms but method expect s
-			int ms = datasource.getNetworkTimeout();
-			int s = DEFAULT_TIMEOUT;
-			if (ms > 0) s = (int) Math.ceil(ms / 1000);
-
+			int s = datasource.getNetworkTimeout();
+			//if timeout is unreasonable, use default
+			if (s < 1 || s > 60) s = DEFAULT_TIMEOUT;	
 			return conn.isValid(s) ? Boolean.TRUE : Boolean.FALSE;
```
michaeloffner, line 220: we no longer convert from milliseconds to seconds (`ms / 1000`), is that okay?

darrell-rg: If you look at the source for [getNetworkTimeout()](https://github.com/lucee/Lucee/blob/5.0/core/src/main/java/lucee/runtime/db/DataSourceSupport.java#L245)  you see it is actually hardcoded to return the constant NETWORK_TIMEOUT_IN_SECONDS ,  which is set to 10.   So the only implementation I could find in the code base was not actually returning milliseconds.  


Also, Math.ceil(10/1000) outputs zero due to using integer division,  so the conversion was never working right anyway.    
image

linsomniac:
Just an additional datapoint: We've been running this fix in our staging environment for 2-3 weeks, and it resolved the problems we were having under 5.3.9 and 5.3.10 with needing lucee restarts after we do database reloads weekly or have done database restarts to test it.

PR #1928, opened 2023-01-12
Update de.xml

andreasRu:
This PR adds translations for debugging-section and bundle-section that were missing, or were defined in English. The git diff shows some data being deleted, but this is not the case: It's just the order of the data that confuses git diff.

For these translations I'm using already a tool, that I've created/developed and can be seen here:
https://github.com/andreasRu/lucee-admin-language-editor

This tool greatly improves speed of translations.


andreasRu:
Need to do some further checks on the resulting values.

PR #1936, opened 2023-01-19
updated a test to Array Slice Implementation for LDEV-1813

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-1813
https://luceeserver.atlassian.net/browse/LDEV-4372


zspitzer:
this is passing coz lucee can't compile the syntax and thus the build ignores the test

PR #1956, opened 2023-02-13
Added a fix to schedule task page is blank in single mode for LDEV-3347

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3347

### DIFF
--- a/core/src/main/cfml/context/admin/index.cfm
+++ b/core/src/main/cfml/context/admin/index.cfm
@@ -3,6 +3,6 @@
 		include "web.cfm";
 	}
 	else {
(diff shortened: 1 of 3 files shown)


zspitzer:
rather than adding reinit to the urls, i'd suggest just removing `application.languages` here, or calling `applicationStop()`

https://github.com/lucee/Lucee/blob/6.0/core/src/main/cfml/context/admin/overview.cfm#L13 

which will have the same effect, but only once, as it's triggered here

https://github.com/lucee/Lucee/blob/6.0/core/src/main/cfml/context/admin/resources/text.cfm#L14

PR #1973, opened 2023-03-02
LDEV-4410  stop queryparam casting empty strings to nulls

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4410


mchupp:
This is a breaking error with no remedy and requires recoding. What issue did it fix?

zspitzer:
@mchupp Sorry, I've updated the usage note on Lucee docs and ticket linked above, with info on how to override this behaviour. There is a link on the ticket to the mailing list discussion which explains why the change was proposed and made.

TLDR, it's for ACF compat and less error prone that the old behaviour

https://docs.lucee.org/reference/tags/queryparam.html#usage-notes

PR #1980, opened 2023-03-14
Added a fix to verify the cache extension before check the cache count for LDEV-4422

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4422

**Memcached extension** is not supported in  **cacheCount()** function.
https://docs.lucee.org/guides/extensions/extension-membase.html


zspitzer:
Um, how come the memcached extension is hard wired in this PR? the cache page handles different cache types?

PR #1993, opened 2023-03-30
LDEV-3845 support utf8 email addresses

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-3845


zspitzer:
We can't write a test for this using our CI as greenmail doesn't support SMTPUTF8
https://github.com/greenmail-mail-test/greenmail/issues/535

PR #1994, opened 2023-03-30
Added a testcase to Check decodeFromURL() for LDEV-4437

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4437


zspitzer:
@cfmitrah can you re-file against the esapi-extension? https://github.com/lucee/extension-esapi/tree/master/tests

PR #1995, opened 2023-03-31
Added a fix to correctly throw an error message using invalid component in the event gateway for LDEV-4438

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4438


zspitzer:
@cfmitrah let's prefix that error with Gateway component, so make it a bit clearer?

PR #2001, opened 2023-04-06
Revert the fix Backward incompatibility (ACF incompatibility) for LDEV-4306 (5.4)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4306


michaeloffner:
@cfmitrah the underlaying ticket is already set to resolved and for Lucee 5.4 we only do security and crucial updates, feel free to reopen if that PR matches the criteria

PR #2005, opened 2023-04-11
Fix error handling and HSQLDB fallback

bdw429s:
Since Java Streams (or lambdas, more specifically) require any exceptions to be unchecked, the `RuntimeException` was bypassing our HSQLDB fallback.  The fix is to fix out the root cause to decide when to fall back to HSQLDB as before,


bdw429s:
@zspitzer

PR #2015, opened 2023-04-19
java: fix typos

jbampton:


zspitzer:
@jbampton the loader has a fixed namespace, so i don't think we can make any changes to function names, @michaeloffner ?

michaeloffner:
I'm aware of this issue for a long time, but sadly we cannot simply chnage it, because that would make it necessary for everyone to replace the lucee.jar, what we only do with SOME major release (last time Lucee 4->5). But i will mark this as a future change.

michaeloffner:
added the future mark for this https://github.com/lucee/Lucee/commit/095b0f52fd2ab146e7b9f88bf3b584d9e753b5fe

PR #2020, opened 2023-04-26
LDEV-4452 switch to maven-resolver-ant-tasks

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4452

/maven-ant-tasks is deprecated and has hard wired references to maven over http, causing new builds without an existing maven cache to fail


zspitzer:
alas none of these tweaks are resolving the problem (build works locally for me)

it can be reliably reproduced via https://github.com/lucee/Lucee/actions/workflows/no-cache.yml

here is an example of migrating from ant maven tasks
https://github.com/ome/bioformats/pull/3923

zspitzer:
@cfmitrah finally got this working (at least with ant!)

zspitzer:
Now mvn is working too

PR #2050, opened 2023-05-15
Added a fix to include missing keys in de.json file for LDEV-4482

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4482


cfmitrah:
The issue is fixed. So I closed this PR

PR #2068, opened 2023-05-22
Added a testcase to collection functions

cfmitrah:


zspitzer:
haha, didn't know these function existed!

PR #2093, opened 2023-06-01
fix http tests

zspitzer:


pfreitag:
@zspitzer FYI https://httpbin.org/ is a great site for testing http clients

PR #2115, opened 2023-06-07
Added a fix to add a sessionCookie settings in GetApplicationSettings() for LDEV-4515 

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4515


zspitzer:
@cfmitrah plz use use https://github.com/lucee/Lucee/blob/6.0/core/src/main/java/lucee/runtime/listener/SessionCookieDataImpl.java#L85

PR #2128, opened 2023-06-14
Added a testcase to check query() for LDEV-4555

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4555


zspitzer:
5.3 is EOL

see https://github.com/lucee/Lucee/pull/2129

PR #2130, opened 2023-06-16
LDEV-4564 when a query is defined with types, don't change

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4564


michaeloffner:
the current behaviour of Lucee is not by accident as it is, this is by design (more details in the ticket).
We can consider a change like this, but then it need to be based on a optional setting made in application.cfc and/or with the query functions.

PR #2133, opened 2023-06-19
Update text.cfm in the Lucee 6 admin 

andreasRu:
This PR ensures to set english values as default for uncompleted resource files.


zspitzer:
To reproduce the error open mappings in the admin in German

PR #2152, opened 2023-06-30
LDEV-4571 admin res 404 after switching to single mode

andreasRu:
https://luceeserver.atlassian.net/browse/LDEV-4571


andreasRu:
Im closing this, after confirming with Zac to do so. This PR is a little bit stinky because it doesn't solve the problem in it's root issue. It works, but it shouldn't be implemented.

PR #2169, opened 2023-07-10
LDEV-1348   Change Function Caching to be Specific to Instance

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-1348

https://github.com/lucee/Lucee/pull/243


zspitzer:
but simple approach that would make cache keys change between restarts… yeah not so useful

PR #2170, opened 2023-07-11
LDEV-4627 (5.4)  - Native QoQ exception when column case differs

bdw429s:


bdw429s:
There are a couple whitespaces changes VSCode threw in automatically, but all this pull is really doing is changing a `Set<String>` variable that collected column names to be a `Set<Key>` so it would automatically avoid duplicates when the columns used different case in different places of the SQL.

bdw429s:
https://luceeserver.atlassian.net/jira/software/c/projects/LDEV/issues/LDEV-4627

PR #2171, opened 2023-07-11
LDEV-4627 (6.0)  - Native QoQ exception when column case differs

bdw429s:
Same pull as https://github.com/lucee/Lucee/pull/2170 but for Lucee 6.0


bdw429s:
https://luceeserver.atlassian.net/jira/software/c/projects/LDEV/issues/LDEV-4627

PR #2184, opened 2023-07-19
Added a fix to Admin bundle detail Page to show the bundle header correctly

cfmitrah:


michaeloffner:
@cfmitrah please confirm that the change does would it should do

PR #2197, opened 2023-07-27
LDEV-4198 timezone issue with dateformat iso

mattdyer:
Depending on your local timezone you could get an incorrect date when using the dateformat 'iso' mask.

https://luceeserver.atlassian.net/browse/LDEV-4198


zspitzer:
thanks matt, 

PS: can you include the link to the ticket in the first comment and drop the PR link on to the ticket for all Lucee PRs please, just makes life easier for everyone!

PR #2214, opened 2023-08-18
Added a testcase to check SerializeJSON() with CFC instance for LDEV-4676

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4676


zspitzer:
@cfmitrah please rebase and enable

PR #2221, opened 2023-08-30
Added testcase for SanitizeHtml() function.

cfmitrah:


zspitzer:
@cfmitrah please refile against the esapi extension

PR #2227, opened 2023-09-08
LDEV-4648 support partially escaped sequence

hideo67:
This is the fix for the issue discussed in

- https://dev.lucee.org/t/non-utf-8-form-data-decoded-incorrectly/12818/18
- https://luceeserver.atlassian.net/jira/software/c/projects/LDEV/issues/LDEV-4648


michaeloffner:
i have 2 concern here

1. some characters are using more than one byte, specially emojis.
2.  if we do any change here, we should switch to apache commons instead

hideo67:
I merged 6.1 into this PR branch, and changed the implementation to use apache commons.

zspitzer:
Can you refile this the PR, since the merges it's become impossibly confusing to review?

hideo67:
I'm closing this since I refiled it as #2335

PR #2233, opened 2023-09-15
Added a testcase in QoQ IN-operator for LDEV-4695

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4695


zspitzer:
@cfmitrah can you review why this failed? maybe just rebase initially?

PR #2239, opened 2023-09-18
sync driver init

davidAtInleague:
The method here (`DataSourceSupport.initialize(Config config)`) is also on the stack as part of https://luceeserver.atlassian.net/browse/LDEV-4701. I can definitely stop 2+ threads on the assignment `driver = _initializeDriver(cd, config);` statement, so it's a bit racy. However, I'm not sure what the consequences are if one thread goes off and does work with driver A, and other threads go off and do work with driver B, so some Datasource D has 2 separate driver instances "associated" with it. I imagine it would be harder to track down than https://github.com/lucee/Lucee/commit/fe6c1792b2ab20dafe85d540d01f9056b8d60752 because it won't result in a NullPointerException but something else more subtle.


zspitzer:
fixed via https://github.com/lucee/Lucee/commit/fe6c1792b2ab20dafe85d540d01f9056b8d60752

PR #2248, opened 2023-09-27
Added a fix to wrap the mapping physical path inside a container for LDEV-4496(5.4)

cfmitrah:


michaeloffner:
@cfmitrah great work on this, but we only do securty and urgent fixes on Lucee 5. please reopen PR if that is the case here

PR #2254, opened 2023-09-29
Correct typos in some admin messages

carehart:
I found one typo, and as long as I was changing it I tried to find more in the file and found a few. That said, I can't confirm I checked the entire file but about half of it.

Finally, there is in the repo an xml version of the file which has the same errors, test/components/Administrator/en.xml. Since that has test in its path, I'll assume it's not relevant/needing to be changed.


zspitzer:
cheers!

PR #2257, opened 2023-10-12
Added a fix to deserializeJSON() throw error for empty String in LDEV-3413 (5.4)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3413


michaeloffner:
we only do security updates on LTS versions

PR #2258, opened 2023-10-13
Added a fix to works correctly cfheader/cfparam tag with end tag for LDEV-4428 (5.4)

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4428


michaeloffner:
@cfmitrah the underlaying ticket is already set to resolved. we no longer add non security fixes to Lucee 5

PR #2305, opened 2024-02-01
Improved the error message in s3 for LDEV-4718

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-4718


michaeloffner:
We already have this fix for Lucee 6, Lucee 5 is in LTS mode and only get crucial security updates

PR #1758, opened 2022-08-13
LDEV-2127 use stringbuilder for bytecode

zspitzer:
from  https://github.com/brucekirkpatrick/Lucee/commit/474ecd9f03e64b4c5741d72cb7955e7ca5bc44f7

https://luceeserver.atlassian.net/browse/LDEV-2127

PR #2315, opened 2024-02-20
LDEV-2127 use stringbuilder for bytecode (v2)

zspitzer:
refiled from https://github.com/lucee/Lucee/pull/1758 as rebase got ugly)

https://luceeserver.atlassian.net/browse/LDEV-2127

code review comments still pending

### DIFF
--- a/core/src/main/java/lucee/transformer/bytecode/op/OpString.java
+++ b/core/src/main/java/lucee/transformer/bytecode/op/OpString.java
@@ -18,7 +18,10 @@
  */
 package lucee.transformer.bytecode.op;
 
+import lucee.transformer.bytecode.literal.LitStringImpl;
+import lucee.transformer.bytecode.cast.CastString;
 import org.objectweb.asm.Type;
+import org.objectweb.asm.commons.GeneratorAdapter;
 import org.objectweb.asm.commons.Method;
 
 import lucee.transformer.TransformerException;
@@ -33,9 +36,15 @@ public final class OpString extends ExpressionBase implements ExprString {
 
 	private ExprString right;
 	private ExprString left;
-
+	
 	// String concat (String)
 	private final static Method METHOD_CONCAT = new Method("concat", Types.STRING, new Type[] { Types.STRING });
+	private static final Method CONSTR_NO_STRING = new Method("<init>", Types.VOID, new Type[] { });
+	private static final Method CONSTR_STRING = new Method("<init>", Types.VOID, new Type[] { Types.STRING });
+	private static final Method APPEND_OBJECT = new Method("append", Types.STRING_BUILDER, new Type[] { Types.OBJECT });
+	private static final Method APPEND_STRING = new Method("append", Types.STRING_BUILDER, new Type[] { Types.STRING });
+	private static final Method TO_STRING = new Method("toString", Types.STRING, new Type[] {});
+
 	private static final int MAX_SIZE = 65535;
 
 	private OpString(Expression left, Expression right) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/lucee/transformer/bytecode/op/OpString.java
```diff
@@ -55,9 +64,40 @@ public static ExprString toExprString(Expression left, Expression right, boolean
...
-		left.writeOut(bc, MODE_REF);
-		right.writeOut(bc, MODE_REF);
-		bc.getAdapter().invokeVirtual(Types.STRING, METHOD_CONCAT);
+		GeneratorAdapter adapter=bc.getAdapter();
+		// new bytecode generation that does the same optimization for string concatenation that the Java compiler does.
+
+		// only creates the stringbuilder object once
+		adapter.newInstance(Types.STRING_BUILDER);
+		adapter.dup();
+		adapter.invokeConstructor(Types.STRING_BUILDER, CONSTR_NO_STRING); // TODO pass left into constructor
```
michaeloffner, line 73: instead of calling append twice, call the constructor with a string, so instead of 
`new StringBuilder().append(left).append(right).toString();`
you do 
`new StringBuilder(left).append(right).toString();`

OR maybe the following is faster?
`new StringBuilder(left.length()+right.length()).append(left).append(right).toString();`

michaeloffner: i still think that StringBuilder comes with unnecessary overhead

michaeloffner: thinking about it, i think we can break down what StringBuilder makes and create a method that does the same but much more optimized

michaeloffner: created some test code 
```

public static void main(String[] args) {
		int max = 1000000;
		double tmp, start, res1 = Double.MAX_VALUE, res2 = Double.MAX_VALUE, res3 = Double.MAX_VALUE, res4 = Double.MAX_VALUE;
		String left = "John Maynard war unser Steuermann, Aushielt er, bis er das Ufer gewann, ";
		String right = "Er hat uns gerettet, er trägt die Kron, Er starb für uns, unsre Liebe sein Lohn.";
		for (int y = 0; y < 100; y++) {

			// ()
			start = SystemUtil.millis();
			for (int i = 0; i < max; i++) {
				new StringBuilder().append(left).append(right).toString();
			}
			tmp = SystemUtil.millis() - start;
			if (tmp < res1) res1 = tmp;

			// (String)
			start = SystemUtil.millis();
			for (int i = 0; i < max; i++) {
				new StringBuilder(left).append(right).toString();
			}
			tmp = SystemUtil.millis() - start;
			if (tmp < res2) res2 = tmp;

			// (int)
			start = SystemUtil.millis();
			for (int i = 0; i < max; i++) {
				new StringBuilder(left.length() + right.length()).append(left).append(right).toString();
			}
			tmp = SystemUtil.millis() - start;
			if (tmp < res3) res3 = tmp;

			// concat
			start = SystemUtil.millis();
			for (int i = 0; i < max; i++) {
				left.concat(right);
			}
			tmp = SystemUtil.millis() - start;
			if (tmp < res4) res4 = tmp;

		}
		print.e("<init>():	" + res1);
		print.e("<init>(String):	" + res2);
		print.e("<init>(int):	" + res3);
		print.e("concat:		" + res4);
	}
```

gives me the following result
```
<init>():	13.766375005245209
<init>(String):	12.96975001692772
<init>(int):	11.805583000183105
concat:		17.216666996479034


```

so it is faster, but my open question ism how many instaces of StringBuilder will we have that the GC need to clean up, atm we win 6ms with  1'000'000 concats

PR #2323, opened 2024-02-23
Added a fix for LDEV-3496

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-3496


michaeloffner:
fix is already applied to Lucee 6 https://github.com/lucee/Lucee/commit/3b589be84f5253ce48752b23e2912e83490aee7a and for Lucee 5 we only make security fixes and regression fixes.

PR #2330, opened 2024-03-13
LDEV-4854 use stricter application options for admin (search results, implicit scopes, limit evaluation)

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4854


zspitzer:
related https://github.com/lucee/Lucee/pull/2329

PR #2355, opened 2024-04-23
Support java 11 and 17 build again with 6.1

zspitzer:


zspitzer:
maven still is throwing an error when building with 11

https://github.com/lucee/Lucee/blob/eb17e3bcec4dc9468faad3f64b0e84b6d701be79/loader/pom.xml#L302

PR #2362, opened 2024-05-14
Update the test case for LDEV-2510

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-2510


zspitzer:
@cfmitrah Micha said you need to enable the search subdirectories settings for this to work?

PR #2382, opened 2024-06-07
LDEV-4866 fix query parser out of bounds

zspitzer:
and https://luceeserver.atlassian.net/browse/LDEV-4867


bdw429s:
Thanks Zac.  Since 5.4.5 is an LTS release, it gets bug fixes right?

PR #2383, opened 2024-06-07
update hsqldb to 2.7.3

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-4887


zspitzer:
see https://github.com/lucee/Lucee/pull/2476 instead

PR #2399, opened 2024-07-05
LDEV-4866 fix query parser out of bounds (6.2) take 2

zspitzer:

refiled from https://github.com/lucee/Lucee/pull/2382 against 6.2


core/src/main/java/lucee/commons/lang/ParserString.java
```diff
@@ -700,6 +711,42 @@ public boolean removeSpace() {
...
+					int end = sql.indexOf("*/", i + 2);
+					if (end != -1) {
+						i = end + 1;
+						continue;
+					}
+				}
+
+				// handle single line comment
+				if (c == '-' && i < (sqlLen - 1) && sql.charAt(i + 1) == '-') {
+					int end = sql.indexOf('\n', i + 1);
```
michaeloffner, line 739: does not handle if the comment is at the last line of the SQL

core/src/main/java/lucee/commons/lang/ParserString.java
```diff
@@ -700,6 +711,42 @@ public boolean removeSpace() {
...
+		char c;
+		int sqlLen = sql.length();
+		StringBuilder sb = new StringBuilder();
+
+		for (int i = 0; i < sqlLen; i++) {
+			c = sql.charAt(i);
+			if ( i < (sqlLen - 1)) {
+				// handle multi line comment
+				if (c == '/' && sql.charAt(i + 1) == '*') {
+					int end = sql.indexOf("*/", i + 2);
```
michaeloffner, line 730: you don't handle at all if the multi line or single line comment is within a literal string like
`select '/* multi */  // -- single' as comments`, that actually is the most complicated part about parsing, be aware of the current context ans switch between them. This also reinvents the wheel for code we already have in place.

PR #2404, opened 2024-07-20
LDEV-5034 test cases for FileSetAccessMode and extract( "tgz" )

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5034


zspitzer:
can't test this coz the build is broken, but i tested this on 6 and it worked, bar checking the mode from directoryList which only works in 6.1 onwards.... swings n roundabouts

i'll remove the debug output once i can actually run the tests!

PR #2425, opened 2024-10-30
LDEV-4696 - Feat/allow Lucee threads to cooperatively stop avoiding threadDeath in tomcat

xdecock:
Hello,

This is a rewrite of a previously closed Pull request,

The goal of this mr is to (at the cost of a slight overhead) allow the inLucee() code to cooperate, to help out with readability, i've used a bytecode "generator", which should reduce the clutter quite a bit

This approach won't for obvious reasons solve all the code path possibles (an infinite recursion is out of scope) but it should allow a better way to stop threads dynamically.

This code has been used in production with millions of request per day for years now, and we started see thread death again by trying to go back to vanilla lucee and getting traffic spikes.

If there is anything that can be done to help getting this in a better state, please let me know.

In leads for improvement:
- Use processingDirective to generate non preemptive bytecode to allow better performances (not sure lot of people needs it, but looks doable here).
- If some other code path require to be made preemptive, i'd gladly be pointed there, i've updated the most obvious paths.
- If code style need reworks, tell me.

https://luceeserver.atlassian.net/browse/LDEV-4696

### DIFF
--- a/core/src/main/java/lucee/transformer/bytecode/statement/While.java
+++ b/core/src/main/java/lucee/transformer/bytecode/statement/While.java
@@ -75,7 +75,7 @@ public final class While extends StatementBaseNoFinal implements FlowControlBrea
 		adapter.ifZCmp(Opcodes.IFEQ, end);
 
 		body.writeOut(bc);
-		InterruptHandlerInjector.writeLoopBodyEnd(adapter, loopCounter, begin, "during for loop");
+		InterruptHandlerInjector.writeLoopBodyEnd(adapter, loopCounter, begin, "during while loop");
 
 		adapter.visitLabel(end);
 	}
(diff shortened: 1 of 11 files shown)


core/src/main/java/lucee/transformer/bytecode/statement/While.java
```diff
@@ -67,13 +68,14 @@ public While(boolean b, Body body, Position start, Position end, String label) {
...
 		GeneratorAdapter adapter = bc.getAdapter();
+		final int loopCounter = InterruptHandlerInjector.writeLoopInit(adapter);
 		adapter.visitLabel(begin);
 
 		expr.writeOut(bc, Expression.MODE_VALUE);
 		adapter.ifZCmp(Opcodes.IFEQ, end);
 
 		body.writeOut(bc);
-		adapter.visitJumpInsn(Opcodes.GOTO, begin);
+		InterruptHandlerInjector.writeLoopBodyEnd(adapter, loopCounter, begin, "during for loop");
```
xdecock:
```suggestion
		InterruptHandlerInjector.writeLoopBodyEnd(adapter, loopCounter, begin, "during while loop");
```

PR #2495, opened 2025-01-31
LDEV-2954 don't create j2ee session for threads

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-2954


zspitzer:
fixed via https://github.com/lucee/Lucee/pull/2492

PR #2498, opened 2025-02-03
Updated and added the fix for LDEV-5259

cfmitrah:
https://luceeserver.atlassian.net/issues/LDEV-5259


zspitzer:
@cfmitrah isn't the reinit required to load the new extension / plugin? (only the logout isn't required)

PR #2512, opened 2025-02-12
add more lucee system constants

zspitzer:


michaeloffner:
this completely break the main purpose of the KeyConstants, the possibility to  hard link them in compiled files!

PR #2519, opened 2025-03-03
LDEV-5382 GetHTTPTimeString is wrongly using hyphens in its datetime mask

shane-tw:
GetHTTPTimeString uses "EE, dd MMM yyyy HH:mm:ss zz" on both Lucee 6.0 and Adobe CF.
As part of [this date refactor commit](https://github.com/lucee/Lucee/commit/94e0d4af9fcd6085be65fbdc26c93fd95e0f61e3#diff-e5dc03e1198ceaaa9659942fde54e3d8da668d23396646fd65b208462e15094eR273) the mask was accidentally changed to "EE, dd-MMM-yyyy HH:mm:ss zz", making it not match Adobe's behaviour anymore, on Lucee 6.2.

This PR reverts it to the old behaviour, so oldFormat behaves as documented:
https://github.com/lucee/Lucee/blob/2f849ff1eb58f3c9b302dea98d0f2dba635e0789/core/src/main/java/lucee/commons/date/DateTimeUtil.java#L262-L265

### DIFF
--- a/core/src/main/java/lucee/commons/date/DateTimeUtil.java
+++ b/core/src/main/java/lucee/commons/date/DateTimeUtil.java
@@ -34,6 +34,7 @@ import lucee.runtime.type.dt.DateTimeImpl;
 
 public abstract class DateTimeUtil {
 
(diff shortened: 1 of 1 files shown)


zspitzer:
Thanks, can you file a ticket plz

PR #2526, opened 2025-03-21
Added the test cases for the LDEV-5406.

cfmitrah:
https://luceeserver.atlassian.net/issues/LDEV-5406


cfmitrah:
@zspitzer , I updated the test case. Could you please review it?

PR #2534, opened 2025-03-27
Added the fix for ticket LDEV-5446.

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-5446


zspitzer:
@cfmitrah can we avoid onclick and use name="cancel" like elsewhere?

cfmitrah:
@zspitzer , I have modified the name to 'Cancel' and  using the onclick to redirect to the main page, similar to how it's done in the datasource page.

PR #2538, opened 2025-04-01
Updated the testcase for LDEV-5457.

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-5457


zspitzer:
@cfmitrah please always output `e.stacktrace` rather than `e.message`

PR #2542, opened 2025-04-10
Added the testcase for LDEV-5486

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-5486


cfmitrah:
I mistakenly committed my changes to the 6.2 branch instead of 7.0

PR #2566, opened 2025-05-07
Add unit test for LDEV-5568 bug 

dswitzer:


zspitzer:
@dswitzer thanks for the test, please always add a `skip=true` to disable the test, once you have confirmed it's failing, so we can merge with breaking the build! ta

i.e.

https://github.com/lucee/Lucee/commit/06dc73072a0ab9e492fe66436a572b37a1f3887e

dswitzer:
I will try to remember that for the future!

PR #2570, opened 2025-05-13
LDEV-5603 - ensure embeded error template have the style and javascript included

jfrobishow:
Reset setting enablecfoutputonly to ensure the embeded style/js is always included

Anything up the stack could have changed the value of enablecfoutputonly multiple time. Make use of Lucee's enhanced function https://docs.lucee.org/guides/Various/lucee-enhanced-functions.html to ensure it always reset back to false. This is safe as we are at the top of the stack and end of the request lifecycle.


jfrobishow:
Closing to recreate based on lucee/Lucee:6.2

PR #2577, opened 2025-05-26
LDEV-2038 allow interface methods with any to match any return types

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-2038


michaeloffner:
this is not the correct way to address this
you need to make an instanceof instead.
Let say you have a component `C` that extend `B` that extends `A` and the interface accept `A`, you `B` and `C` also should be accepted because they are instanceof `A`, same as `struct` is instanceof `any`

PR #2589, opened 2025-06-10
LDEV-5339: Testcase for isThreadInterrupted()

cfmitrah:
https://luceeserver.atlassian.net/issues/LDEV-5339


zspitzer:
@cfmitrah plz reduce the interval for this test, if 500 works, so should 50ms?

PR #2600, opened 2025-06-26
Update Controler.java

tomchiverton:
Fix typo in log string


zspitzer:
wrong branch. these logs are based on the misnamed class name, so we probably won't change that

tomchiverton:
Oh, sorry, what's the correct current branch (for v6.x) ?
It can't be correct that some log "controller" and some "controler" no matter the Java class name ?

PR #2607, opened 2025-07-08
LDEV-5689 improve Query.cfc performance

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5689


zspitzer:
these changes are going into 7 https://github.com/lucee/Lucee/pull/2609 only

PR #2626, opened 2025-07-18
For LDEV-5653, Initially added the bundleName and bundleVersion fields to the JDBC Other datasource form.

cfmitrah:
https://luceeserver.atlassian.net/browse/LDEV-5653


cfmitrah:
The changes have been updated in a new branch.

PR #2629, opened 2025-07-24
LDEV-5713 update various commons libs (lang3,text,io,compress)

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5713


michaeloffner:
do not merge this unless you are certain this does not cause any version conflicts with any library also reference this libraries, including ALL extensions.

PR #2638, opened 2025-08-02
LDEV-5744 fix bash date parsing, ignore repeated whitespace

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5744

### DIFF
--- a/core/src/main/java/lucee/runtime/op/date/DateCaster.java
+++ b/core/src/main/java/lucee/runtime/op/date/DateCaster.java
@@ -160,6 +160,7 @@ public final class DateCaster {
 		timeZone = ThreadLocalPageContext.getTimeZone(timeZone);
 		DateTime dt = toDateSimple(str, convertingType, true, timeZone, defaultValue);
 		if (dt == null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/lucee/runtime/op/date/DateCaster.java
```diff
@@ -160,6 +160,7 @@ public static DateTime toDateAdvanced(String str, short convertingType, TimeZone
 		timeZone = ThreadLocalPageContext.getTimeZone(timeZone);
 		DateTime dt = toDateSimple(str, convertingType, true, timeZone, defaultValue);
 		if (dt == null) {
+			str =StringUtil.suppressWhiteSpace(str);
```
michaeloffner, line 163: this looks like a possible regression to me, are you certain there is no case when this whitespace is needed?

zspitzer: Well, none of the date masks in formatUtil have multiple spaces? bash does have two

michaeloffner:
manually adapted

PR #2648, opened 2025-08-25
LDEV-5771 add supported random algorithms to rand exception

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5771

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/math/Rand.java
+++ b/core/src/main/java/lucee/runtime/functions/math/Rand.java
@@ -69,9 +69,12 @@ public final class Rand implements Function {
 					result = SecureRandom.getInstance(algorithm);
 				}
 				catch (NoSuchAlgorithmException e) {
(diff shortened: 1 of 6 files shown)


pfreitag:
@zspitzer - you know what would be cool, if you could use algorithm "SecureRandom" and it just used `SecureRandom.getInstance()` or "SecureRandomStrong" and it used `SecureRandom.getInstanceStrong()`, instead of having to specify an algorithm. I believe the default choice can vary by OS, and this way your code isn't tied to a particular algorithm (if for example a weakness is discovered in the algorithm, you can just expect that updating the JVM will fix it).

It is kind of like saying I want to explicitly use TLS 1.2 for https because it is the best, then TLS 1.3 comes out and you have to update all your code. It typically ends up being better to just use the JVM default because then your code can update automatically when TLS 1.4 comes out, etc.

zspitzer:
Agreed, good point, we have a ticket about changing the default from CFMX_COMPAT for 7

https://luceeserver.atlassian.net/browse/LDEV-5195?focusedCommentId=64013

PR #2660, opened 2025-09-19
LDEV-5821 optimise cfproperty approach

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5821

### DIFF
--- a/core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
+++ b/core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
@@ -1,0 +1,210 @@
+/**
+ * Copyright (c) 2025, Lucee Association Switzerland. All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
+package lucee.transformer.bytecode.statement.tag;
+
+import org.objectweb.asm.Type;
+import org.objectweb.asm.commons.GeneratorAdapter;
+import org.objectweb.asm.commons.Method;
+
+import lucee.runtime.op.Caster;
+import lucee.transformer.Factory;
+import lucee.transformer.Position;
+import lucee.transformer.TransformerException;
+import lucee.transformer.bytecode.BytecodeContext;
+import lucee.transformer.bytecode.statement.FlowControlFinal;
+import lucee.transformer.bytecode.util.ASMConstants;
+import lucee.transformer.bytecode.util.Types;
+import lucee.transformer.expression.Expression;
+import lucee.transformer.statement.tag.Attribute;
+import lucee.transformer.statement.tag.Tag;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Optimized bytecode generation for cfproperty tags.
+ * Bypasses tag lifecycle overhead by calling ComponentUtil.registerProperty directly.
+ */
+public final class TagProperty extends TagBase {
+
+	private static final Method REGISTER_PROPERTY = new Method("registerProperty", Type.VOID_TYPE, new Type[] {
+		Types.PAGE_CONTEXT, Types.STRING, Types.STRING, Types.OBJECT,
+		Types.STRING, Types.STRING, Types.STRING, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE
+	});
+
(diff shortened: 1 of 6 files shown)


core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+				String val = getLiteralString(attr);
+				setter = !"no".equalsIgnoreCase(val) && !"false".equalsIgnoreCase(val);
+			}
+			else if ("getter".equals(attrName)) {
+				String val = getLiteralString(attr);
+				getter = !"no".equalsIgnoreCase(val) && !"false".equalsIgnoreCase(val);
+			}
+		}
+
+		System.err.println("[DEBUG-PROPERTY] Generating optimized call: ComponentUtil.registerProperty(pc, name=\"" + name + "\", type=\"" + type + "\")");
```
michaeloffner: please remove this.
Tip: use instead print.e or print.o, it is not only better for showing cfml specific tyes, you can lso simply remove the "extends aprint" in that class and you see where it is used and clean up add the end of development.

core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+			}
+			else if ("hint".equals(attrName)) {
+				hint = getLiteralString(attr);
+			}
+			else if ("displayname".equals(attrName)) {
+				displayname = getLiteralString(attr);
+			}
+			else if ("required".equals(attrName)) {
+				String val = getLiteralString(attr);
+				required = "yes".equalsIgnoreCase(val) || "true".equalsIgnoreCase(val);
```
michaeloffner: do  `required=Caster.toBoolean(val,false)` instead.

core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+			}
+			else if ("setter".equals(attrName)) {
+				String val = getLiteralString(attr);
+				setter = !"no".equalsIgnoreCase(val) && !"false".equalsIgnoreCase(val);
```
michaeloffner: that is a bit strange i would assume you wanna do the same as above, anyway we never check for "false" and "no" or "true" or "yes" in the wild. the only please that is done is within Caster.toBoolean().

core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+			}
+			else if ("getter".equals(attrName)) {
+				String val = getLiteralString(attr);
+				getter = !"no".equalsIgnoreCase(val) && !"false".equalsIgnoreCase(val);
```
michaeloffner: dito above

core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+
+		// setter parameter
+		adapter.push(setter);
+
+		// getter parameter
+		adapter.push(getter);
+
+		// Call ComponentUtil.registerProperty
+		Method method = new Method("registerProperty", Type.VOID_TYPE, new Type[] { Types.PAGE_CONTEXT, Types.STRING, Types.STRING, Types.OBJECT,
+				Types.STRING, Types.STRING, Types.STRING, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE });
```
michaeloffner: i prefer to define Methods as static Constants in the beginning of the class and then just use the constants, that way he class get created only once and not every single time. Alos it makes it easier to read.
```
private static Method REGISTER_PROPERTY = new Method("registerProperty", Type.VOID_TYPE, new Type[] { Types.PAGE_CONTEXT, Types.STRING, Types.STRING, Types.OBJECT,
				Types.STRING, Types.STRING, Types.STRING, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE, Type.BOOLEAN_TYPE });
...
adapter.invokeStatic(Type.getType("Llucee/runtime/type/util/ComponentUtil;"), REGISTER_PROPERTY);
```

core/src/main/java/lucee/transformer/bytecode/statement/tag/TagProperty.java
```diff
@@ -0,0 +1,195 @@
...
+
+		adapter.invokeStatic(Type.getType("Llucee/runtime/type/util/ComponentUtil;"), method);
+
+		System.err.println("[DEBUG-PROPERTY] TagProperty optimization complete");
```
michaeloffner: remove

core/src/main/java/lucee/transformer/cfml/script/AbstrCFMLScriptTransformer.java
```diff
@@ -1644,7 +1644,14 @@ private final Tag _propertyStatement(Data data, Body parent) throws TemplateExce
 		Position line = data.srcCode.getPosition();
 
 		TagLibTag tlt = CFMLTransformer.getTLT(data.srcCode, "property", data.config.getIdentification());
-		Tag property = new TagOther(data.factory, line, null);
+		Tag property;
+		try {
+			property = tlt.getTag(data.factory, line, null);
+		}
+		catch (Exception e) {
+			// Fall back to TagOther if TTT instantiation fails
```
michaeloffner, line 1653: we should log this in the template log

michaeloffner: like this LogUtil.log("template", "property",null);

PR #2666, opened 2025-09-22
LDEV-5832 improve source code parsing and compiler performance

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5832

### DIFF
--- a/core/src/main/java/lucee/transformer/bytecode/PageImpl.java
+++ b/core/src/main/java/lucee/transformer/bytecode/PageImpl.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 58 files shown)


core/src/main/java/lucee/transformer/bytecode/PageImpl.java
```diff
@@ -913,7 +914,7 @@ public boolean writeLog() {
 		return _writeLog && !isInterface();
 	}
 
-	public static void registerFields(BytecodeContext bc, List<LitString> keys) throws TransformerException {
+	public static void registerFields(BytecodeContext bc, LinkedHashMap<LitString, Integer> keys) throws TransformerException {
```
michaeloffner: is it needed to use the LinkedHashMap class here, can it not be just List as before.
i always try to stck with the interfaces and never the concrete class if not needed.
gives more freedom for the future

like
` public static void registerFields(BytecodeContext bc, Map<LitString, Integer> keys) throws TransformerException {`

core/src/main/java/lucee/transformer/bytecode/PageImpl.java
```diff
@@ -314,7 +315,7 @@ public byte[] execute(String className) throws TransformerException {
 		// not exists in any case, so every usage must have a plan b for not existence
 		PageSource optionalPS = sourceCode instanceof PageSourceCode ? ((PageSourceCode) sourceCode).getPageSource() : null;
 
-		List<LitString> keys = new ArrayList<LitString>();
+		LinkedHashMap<LitString, Integer> keys = new LinkedHashMap<LitString, Integer>();
```
michaeloffner: is it needed to use the LinkedHashMap class here, can it not be just List as before.
i always try to stck with the interfaces and never the concrete class if not needed.
like
`Map<LitString, Integer> keys = new LinkedHashMap<>();`

core/src/main/java/lucee/transformer/bytecode/PageImpl.java
zspitzer: fixed

core/src/main/java/lucee/transformer/bytecode/PageImpl.java
zspitzer: fixed

PR #2679, opened 2025-10-12
LDEV-5690 Allow safe variable dereferencing with limitEvaluation

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5690

### DIFF
--- a/core/src/main/java/lucee/runtime/interpreter/ref/var/Variable.java
+++ b/core/src/main/java/lucee/runtime/interpreter/ref/var/Variable.java
@@ -25,6 +25,8 @@ import lucee.runtime.interpreter.ref.Ref;
 import lucee.runtime.interpreter.ref.RefSupport;
 import lucee.runtime.interpreter.ref.Set;
 import lucee.runtime.interpreter.ref.literal.LString;
+import lucee.runtime.interpreter.ref.func.BIFCall;
+import lucee.runtime.interpreter.ref.func.UDFCall;
 import lucee.runtime.op.Caster;
 import lucee.runtime.type.KeyImpl;
 import lucee.runtime.type.Query;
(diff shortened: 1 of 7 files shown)


core/src/main/java/lucee/runtime/interpreter/ref/var/Variable.java
```diff
@@ -115,4 +117,29 @@ public String getKeyAsString(PageContext pc) throws PageException {
...
+
+		// Block function calls
+		if (ref instanceof BIFCall || ref instanceof UDFCall) {
+			throw new SecurityInterpreterException( "invalid syntax, function calls are not supported." );
+		}
+
+		// Block assignments (Assign and DynAssign are in the same package)
+		String className = ref.getClass().getSimpleName();
+		if ("Assign".equals( className ) || "DynAssign".equals( className )) {
+			throw new SecurityInterpreterException( "invalid syntax, assignments are not supported." );
```
michaeloffner, line 139: could we add here info from where this limitation comes from and how it can be changed?

core/src/main/java/lucee/runtime/interpreter/ref/var/Variable.java
```diff
@@ -115,4 +117,29 @@ public String getKeyAsString(PageContext pc) throws PageException {
...
+	 * @param ref the Ref to check
+	 * @throws SecurityInterpreterException if the Ref contains unsafe operations
+	 */
+	private void checkSecurity(Ref ref) throws SecurityInterpreterException {
+		if (ref == null) return;
+
+		// Block function calls
+		if (ref instanceof BIFCall || ref instanceof UDFCall) {
+			throw new SecurityInterpreterException( "invalid syntax, function calls are not supported." );
+		}
```
michaeloffner, line 134: could we add here info from where this limitation comes from and how it can be changed?

PR #2685, opened 2025-10-24
LDEV-5865 use Object[] for args and local scopes

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5865

### DIFF
--- a/core/src/main/java/lucee/runtime/type/scope/ArgumentImpl.java
+++ b/core/src/main/java/lucee/runtime/type/scope/ArgumentImpl.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import lucee.commons.io.SystemUtil;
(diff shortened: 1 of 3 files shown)


michaeloffner:
there is one big problem with this code and this is maybe the reason you see a performance increase, the original code uses a  thread safe "ConcurrentHashMap", your code does not and yes argument and local scope can be shared between multiple threads easely, for example with closures used in a parallel function like arrayEach.

in my opinion it is impossible 
that code like this
```
for ( int i = 0; i < localCount; i++ ) {
				if ( localArray[i * 2].equals( key ) ) {
					return localArray[i * 2 + 1];
				}
			}
```
is faster than a map, if so we need to choose a better map.
but as i said, this code is not locked.

yes most maps use "equals" internally as well for key comparsion, because of that we use Collection.Key in lucee and not simply a String.

zspitzer:
https://github.com/lucee/Lucee/blob/7.0/core/src/main/java/lucee/runtime/type/scope/LocalImpl.java#L38

```
	public LocalImpl() {
		// super("local", Scope.SCOPE_LOCAL, Struct.TYPE_SYNC, 4);
		super("local", Scope.SCOPE_LOCAL, Struct.TYPE_REGULAR, localInitialCapacity);
	}
```

https://github.com/lucee/Lucee/blob/7.0/core/src/main/java/lucee/runtime/type/scope/ArgumentImpl.java#L79

```
	public ArgumentImpl() {
		// super("arguments", SCOPE_ARGUMENTS, StructImpl.TYPE_LINKED, 4);
		super("arguments", SCOPE_ARGUMENTS, StructImpl.TYPE_LINKED_NOT_SYNC, argumentInitialCapacity);
	}
```

PR #2689, opened 2025-10-25
LDEV-5871 Add cacheprefix attribute to query

deanmaunder:
https://luceeserver.atlassian.net/browse/LDEV-5871


zspitzer:
@deanmaunder i fixed the error with CI, can you rebase please?

please rename the test without dash and use BDD style (describe/it/expect)

As we have multiple cache providers, like redis, memcached and ehcache, we need to test them all

https://github.com/lucee/Lucee/blob/7.0/test/_setupTestServices.cfc#L244

i.e. https://github.com/lucee/Lucee/blob/7.0/test/tickets/LDEV4342.cfc

see https://github.com/lucee/Lucee/blob/7.0/test/tags/dbInfo.cfc for such an approach

Also the tests don't yet cover `cacheRemove()` or `cacheClear()` which is quite an core important aspect for this nice new enhancement

test/tickets/LDEV-5871.cfc
```diff
@@ -0,0 +1,345 @@
...
+		```
+		<cfquery datasource="#dsn#" name="local.result2" cacheprefix="with-params-" cachedwithin="#createTimeSpan(0,0,1,0)#" >
+			SELECT 
+			<cfqueryparam value="test-value" cfsqltype="cf_sql_varchar"> as col, NOW() as executed
+		</cfquery>
+		```
+		assertEquals(result1.executed, result2.executed, "Query with params should be cached");
+	}
+
+	private function repeatString(required string str, required numeric times) {
```
zspitzer: use the BIF ? https://docs.lucee.org/reference/functions/repeatstring.html

zspitzer:
AFAIK This is still currently O(n) without additional changes?

deanmaunder:
> AFAIK This is still currently O(n) without additional changes?

Unsure what you are asking about?

deanmaunder:
@zspitzer anything further need to be done on this?

deanmaunder:
Sorry @zspitzer this one has become critical in our environment, can we get it merged?  Will it backport to 6?  we arent ready for 7 yet.
Otherwise I can build a custom Lucee version and deploy.

deanmaunder:
> AFAIK This is still currently O(n) without additional changes?

You're right that key iteration is still O(n) with the default CacheSupport.remove(CacheKeyFilter) impl. What this PR delivers is skipping value deserialization on cacheClear:

- cacheClear(tags, "query") → QueryTagFilter (CacheEntryFilter) → ce.getValue() on every entry → full deserialization
- cacheClear("prefix:*", "query") → WildCardFilter (CacheKeyFilter) → keys-only scan, no value reads

The issue we were trying to solve was the storm that happened during a cacheClear on Redis where it would read the value of hundreds of large query objects from the cache, just to get the tags.

PR #2693, opened 2025-10-29
LDEV-5643 LDEV-5881 switch to using Java 25 ScopedValues

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5881
https://luceeserver.atlassian.net/browse/LDEV-5643

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
       LUCEE_BUILD_JAVA_VERSION:
         required: true
         type: string
-        default: '21'
+        default: '25'
   push:
     branches:
       - '**' # thus ignoring tagging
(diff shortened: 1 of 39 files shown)


zspitzer:
re: Removal of `InheritableThreadLocal`

Support During the migration from ThreadLocal to ScopedValue, the pcThreadLocalInheritable (InheritableThreadLocal) was removed because ScopedValue does not support automatic inheritance to child threads like InheritableThreadLocal did. 

The INHERIT_ENABLED flag was set to false, and inheritance functionality was intentionally disabled. As part of the performance optimization cleanup, we removed the now-dead "pc from parent thread" code blocks that were checking pcThreadLocalInheritable.get() - these blocks had been reduced to pc = null; 

if (pc != null) after the variable was removed, making them unreachable dead code. Child threads that need access to a parent's PageContext must now explicitly pass it or establish a new ScopedValue scope using ScopedValue.where(ThreadLocalPageContext.CURRENT, pc).call(() -> {}). This is a more explicit and safer pattern for virtual threads.

PR #2694, opened 2025-11-02
LDEV-5887 add LUCEE_PRECISE_MATH=disabled

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5887


core/src/main/java/lucee/runtime/config/ConfigImpl.java
```diff
@@ -6121,6 +6121,9 @@ public boolean getPreciseMath() {
 				if (preciseMath == null) {
 					boolean pm = false;
 					String str = SystemUtil.getSystemPropOrEnvVar("lucee.precise.math", null);
+					// If set to "disabled", ignore it here - ThreadLocalPageContext handles the global kill switch
+					if ("disabled".equalsIgnoreCase(str)) str = null;
+
```
michaeloffner, line 6128: why is this not simply a boolean we never use the word "disabled" for any flag, should simply be a boolean.
also there is no trim to remove whitespaces.
should be 
```
Boolean b = Caster.toBoolean(SystemUtil.getSystemPropOrEnvVar("lucee.precise.math", null),null);
if(b!=null) {...}
else {str = ConfigFactoryImpl.getAttr(root, "preciseMath");...}

File: core/src/main/java/lucee/runtime/engine/ThreadLocalPageContext.java [hunk 2]
```diff
@@ -47,6 +47,14 @@ public final class ThreadLocalPageContext {
...
 	private static ThreadLocal<Boolean> insideInheritableRegistration = new ThreadLocal<Boolean>();
 
+	/**
+	 * Global kill switch for precise math.
+	 * - null: normal behavior, per-application/config control
+	 * - "disabled": globally disabled, throws error if anyone tries to enable it
+	 */
+	private static final String PRECISE_MATH_GLOBAL = SystemUtil.getSystemPropOrEnvVar( "lucee.precise.math", null );
+	public static final boolean PRECISE_MATH_GLOBALLY_DISABLED = "disabled".equalsIgnoreCase( PRECISE_MATH_GLOBAL );
+
```
michaeloffner: same her, just one and a boolean and not "disabled", this is also the wrong place to do this. ThreadLocal classes should only be used for thread local related stuff.

core/src/main/java/lucee/runtime/engine/ThreadLocalPageContext.java
```diff
@@ -124,7 +132,14 @@ public static Config getConfig(Config config) {
 		return config;
 	}
 
+	public static boolean isPreciseMathGloballyDisabled() {
+		return PRECISE_MATH_GLOBALLY_DISABLED;
+	}
```
michaeloffner: this is a new pattern i really not wanna indroduce? having a getter "isWhateverGlobally" and  "isWhatever" and specially not "Disabled" what inverse the logic

core/src/main/java/lucee/runtime/engine/ThreadLocalPageContext.java
```diff
@@ -124,7 +132,14 @@ public static Config getConfig(Config config) {
...
+
 	public static boolean preciseMath(PageContext pc) {
+		// Global kill switch - when set to "disabled", completely bypass all precise math
+		if (PRECISE_MATH_GLOBALLY_DISABLED) return false;
+
```
michaeloffner, line 131: with different name that is fine to keep

core/src/main/java/lucee/runtime/listener/ClassicApplicationContext.java
```diff
@@ -1168,12 +1169,17 @@ public void setRegex(Regex regex) {
...
 		if (preciseMath == null) return cp.getPreciseMath();
 		return preciseMath;
 	}
 
 	@Override
 	public void setPreciseMath(boolean preciseMath) {
+		// Check if precise math has been globally disabled
+		if (ThreadLocalPageContext.isPreciseMathGloballyDisabled() && preciseMath) {
+			throw new RuntimeException( "Precise math has been globally disabled via lucee.precise.math=disabled. Cannot enable it per-application." );
+		}
```
michaeloffner, line 1182: i am not certain, but i have the feeling that a cfconfig setting could trigger this, that would mean that certain cfconfig do not work at all, if that enviroment setting is set. that should not happen. breaking a build because of a setting

core/src/main/java/resource/setting/sysprop-envvar.json
```diff
@@ -613,9 +613,9 @@
 	{
 		"sysprop": "lucee.precise.math",
 		"envvar": "LUCEE_PRECISE_MATH",
-		"desc": "A boolean value. If enabled, this improves the accuracy of floating-point calculations but makes them slightly slower",
+		"desc": "Controls precise math behavior. Values: true (enable BigDecimal for accuracy), false (use Double for performance), 'disabled' (global kill switch - completely bypass all precise math checks, throws error if applications try to enable it)",
 		"category": "performance",
```
michaeloffner, line 625: when i am reading this is am still on the fence, we now have a setting you can not overwrite at all at runtime, in any way.
why do we have this for "precise math", but not for "local mode", i feel if we introduce this, we need to have a general pattern in mind for ALL settings and that needed to be implemented in a way, that does not need to add all this changes for a single setting.
IF we do this, we need a way to make this MUCH simpler, this adds a big layer of complexity to the configuration.
with switching to single mode i tried to simplify things, not only for developers, but also for us, this goes in the opposite direction. 
All this is a good proof of concept, but we need to make a general plan for this.
i just did go through Lucee and i have made all env var setting overrule settings made in cfconfig. but now we add again different behaviour for different settings.

zspitzer: I picked precise math here as it's a real hotspot and this static final approach allows the jvm to completely eliminate the runtime overhead

localmode didn't really show up in any jfr profiling

core/src/main/java/lucee/runtime/listener/ClassicApplicationContext.java
zspitzer: Didn't we change the configuration hierarchy to make env var/sys prop always override cfconfig?
https://luceeserver.atlassian.net/browse/LDEV-5642

it's off by default, do you mean breaking our CI build?

PR #2700, opened 2025-11-11
LDEV-5908 ConcurrentHashMap Optimization

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5908


core/src/main/java/lucee/commons/collection/concurrent/ConcurrentHashMapNullSupport.java
michaeloffner, line 1: null support still work the same way?

zspitzer: yes, now with extra test coverage

https://github.com/lucee/Lucee/commit/11160ca606a8d6a936ed155f166bdda9af5be90c

PR #2701, opened 2025-11-11
LDEV-5903 revert to GC-able classloader per cfml template

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5903


DominicWatson:
I tried something very similar with no joy. Just testing this PR out locally and also not seeing a reversion of the behaviour. Giving up for tonight.

zspitzer:
closing, @micstriit fixed this using a different approach

PR #2716, opened 2025-11-27
LDEV-5941 improve sessionCommit() add SessionTouch() - 7.0

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5941

### DIFF
--- a/core/src/main/java/lucee/runtime/functions/system/SessionTouch.java
+++ b/core/src/main/java/lucee/runtime/functions/system/SessionTouch.java
@@ -8,8 +8,8 @@ import lucee.runtime.type.scope.Session;
 import lucee.runtime.type.scope.storage.IKStorageScopeSupport;
 import lucee.runtime.type.scope.storage.StorageScopeImpl;
 
-public final class SessionCommit implements Function {
-	private static final long serialVersionUID = -2243745577257724777L;
+public class SessionTouch implements Function {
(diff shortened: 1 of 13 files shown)


core/src/main/java/lucee/runtime/functions/system/SessionCommit.java
```diff
@@ -4,12 +4,22 @@
...
-		((PageContextImpl) pc).sessionScope().touchAfterRequest(pc);
+	public static String call( PageContext pc ) throws PageException {
+		Session session = ((PageContextImpl) pc).sessionScope();
+		if ( session instanceof IKStorageScopeSupport ) {
+			((IKStorageScopeSupport) session).setDirty();
+		}
+		else if ( session instanceof StorageScopeImpl ) {
+			((StorageScopeImpl) session).setDirty();
+		}
+		session.touchAfterRequest( pc );
```
michaeloffner, line 22: to be honest i don't like to call a listener method like "touchAfterRequest" outside the flow it should be called.
this method is called AFTER the request is finished, what if a implementation in the future clean up the request and remove data used by this request, it will then break after this function is executed. at best this limits what this method can do, at worst it corrupts the request.

michaeloffner:
do we really need this 2 functions? i think `SessionCommit(immediately:true|false)` is all we need.
Also this should be in against Lucee 7.1, not 7.0.

PR #2723, opened 2025-12-04
LDEV-1402 lucee native debugger support

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-1402


zspitzer:
@micstriit loader changes suck i know....

loader/src/main/java/lucee/runtime/PageContext.java
```diff
@@ -981,8 +981,12 @@ public abstract class PageContext extends jakarta.servlet.jsp.PageContext {
 
 	public abstract void exeLogStart(int position, String id);
 
+	public abstract void exeLogStart(int position, int line, String id);
```
michaeloffner: we cannot update an existing interface in the loader.
providing the line makes no sense, this holds no additional information the position not already provides.

core/src/main/java/lucee/runtime/engine/CFMLEngineImpl.java
```diff
@@ -224,8 +224,24 @@ public final class CFMLEngineImpl implements CFMLEngine {
...
+			CONSOLE_OUT = new lucee.runtime.debug.DebuggerPrintStream(System.out, false);
+			CONSOLE_ERR = new lucee.runtime.debug.DebuggerPrintStream(System.err, true);
+			System.setOut(CONSOLE_OUT);
+			System.setErr(CONSOLE_ERR);
+		}
+		else {
+			CONSOLE_OUT = System.out;
+			CONSOLE_ERR = System.err;
+		}
+	}
```
michaeloffner, line 277: does this work with the existing configuration for system err and system out?

core/src/main/java/lucee/runtime/engine/ExecutionLogSupport.java
```diff
@@ -102,6 +102,12 @@ public final void start(int pos, String id) {
 		map.put(id, new Pair(current, pos));
 	}
 
+	@Override
+	public void start(int pos, int line, String id) {
+		// Default: delegate to pos-only version, subclasses can override to use line
```
michaeloffner: pos areeady hold the line information, so i don't see any real benefit in this, may faster but a bit of a mess.
if you wanna have fast access to the line, the you the "Position" class instead.

core/src/main/java/lucee/runtime/functions/decision/IsDebuggerEnabled.java
```diff
@@ -0,0 +1,11 @@
+package lucee.runtime.functions.decision;
+
+import lucee.runtime.PageContext;
+import lucee.runtime.config.ConfigImpl;
+
+public final class IsDebuggerEnabled {
```
michaeloffner: should implement the Function interface or at least BIF

core/src/main/java/lucee/transformer/bytecode/PageImpl.java
```diff
@@ -736,6 +738,52 @@ else if (functions.length <= 10) {
...
+				execLinesAdapter.push(maxLine);
+				execLinesAdapter.invokeStatic(Types.PAGE_CONTEXT_UTIL,
+					new Method("decodeExecutableLines", Type.getType(int[].class), new Type[] { Types.STRING, Type.INT_TYPE }));
+			}
+		}
+		execLinesAdapter.arrayStore(Types.OBJECT);
+
+		execLinesAdapter.returnValue();
+		execLinesAdapter.endMethod();
+
```
michaeloffner, line 810: so we add this to every class when debugging is enabled or not. how much is the overhead with that?

michaeloffner:
make sure this is backward compatible, existing archives containing Lucee bytecode most likely will no longer work with this change.

core/src/main/java/lucee/runtime/engine/CFMLEngineImpl.java
zspitzer: Yes - there are two separate output mechanisms in Lucee:

- Config writers (`config.getOutWriter()`/`config.getErrWriter()`) - used by `SystemOutput()` and most CFML output. These are completely separate from System.out/System.err and are unaffected by the DebuggerPrintStream wrapper.

- System streams (`System.out`/`System.err`) - used by direct Java code like `aprint`, logging, etc. The `DebuggerPrintStream` wraps these to tee output to the debugger.

Code that manipulates the system streams (like `SystemOut.setOut()`/`setErr()` in the test runner) works fine - it saves the current stream (which may be the `DebuggerPrintStream`), replaces it temporarily, then restores it. The wrapper just gets saved and restored along with everything else. The only thing the wrapper does is intercept writes and forward them to both the original stream and the debugger listener - it doesn't change any other behaviour.

core/src/main/java/lucee/transformer/bytecode/PageImpl.java
zspitzer: i made it conditional

PR #2729, opened 2025-12-09
LDEV-5968 Dirty connection state and leak in DatasourceConnectionFactory

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-5968


zspitzer:
also not a problem, closing

PR #2739, opened 2025-12-29
LDEV-6044 use Felix JPMS auto-detection for JDK packages

zspitzer:
Also remove ancient jars now in JDK due to conflicts

- xml-apis 
- w3-dom

temp remove hibernate ext from build, it pulls in xml-apis which now conflicts (needs updating)

https://luceeserver.atlassian.net/browse/LDEV-6044


zspitzer:
The Hibernate ext pulls in this jar too https://github.com/Ortus-Solutions/extension-hibernate/pull/19

PR #2756, opened 2026-02-04
LDEV-6097 avoid stackoverflow with catchBlockImpl and cause chain

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-6097

### DIFF
--- a/core/src/main/java/lucee/runtime/exp/CatchBlockImpl.java
+++ b/core/src/main/java/lucee/runtime/exp/CatchBlockImpl.java
@@ -35,6 +35,7 @@ import lucee.runtime.engine.ThreadLocalPageContext;
 import lucee.runtime.op.Castable;
 import lucee.runtime.op.Caster;
 import lucee.runtime.op.Decision;
+import lucee.runtime.op.Duplicator;
 import lucee.runtime.reflection.Reflector;
 import lucee.runtime.reflection.pairs.MethodInstance;
 import lucee.runtime.type.Collection;
@@ -67,6 +68,8 @@ public final class CatchBlockImpl extends StructImpl implements CatchBlock, Cast
 	public static final Key ADDITIONAL = KeyConstants._additional;
 
 	private final PageException exception;
+	private CatchBlock cachedCause;
+	private boolean causeCached;
 
 	CatchBlockImpl(PageException pe, int level) {
 		if (level < 0) level = 0;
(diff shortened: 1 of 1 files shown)


michaeloffner:
why should we avoid a deep copy?
i try to create a test case for this, but so far with no success, an exception having a cause alone, does not trigger it.
problem must be that a child points to itself or its paren, but when i do that i get an error much earlier
```
at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
 at lucee.commons.lang.ExceptionUtil.unwrap(ExceptionUtil.java:276)
```

michaeloffner:
this is a classic patch covering the underlaying issue, as said i was not able to reproduce this issue, so i need to speculate, but the problem is that the catch block is created new every time.
So we need not to cache the result from  `Caster.toPageException(cause).getCatchBlock(config)`, we need to make sure the PageException does not create a new catch block every time and maybe also make sure that native exception not get created new.
second is more complicated, so we start with the first and see what happens.

PR #2764, opened 2026-03-11
LDEV-6135 Added test case for CFTHREAD JOIN issue

cfmitrah:


zspitzer:
adapted into https://github.com/lucee/Lucee/commit/be1a7304b30ca0bb5c06f9e2db7de550bc291e6d

PR #2767, opened 2026-03-12
LDEV-6144 - recover from stale Felix cache on startup (bundle name conflict)

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-6144

## Summary

- When Felix starts and encounters "Bundle symbolic name and version are not unique" (caused by a stale cache, e.g. from LDEV-6145), Lucee would crash on startup with no recovery
- Also fixes an existing bug in `getFelix()` where "Error creating bundle cache" was caught and the cache cleared, but Felix was never restarted — returning an unstarted Felix instance causing an NPE
- Fix: catch the "not unique" `BundleException` from `BundleLoader.loadBundles()`, clear the stale Felix cache, and retry with a fresh Felix instance

## Why this is needed alongside LDEV-6145

LDEV-6145 prevents future corruption, but users who already have a corrupted Felix cache (from an older Lucee version) cannot start Lucee at all — core never loads, so the LDEV-6145 fix never runs. This loader-level fix recovers from the corrupted state automatically on next restart.

### DIFF
--- a/loader/src/main/java/lucee/loader/engine/CFMLEngineFactory.java
+++ b/loader/src/main/java/lucee/loader/engine/CFMLEngineFactory.java
@@ -675,7 +675,19 @@ public class CFMLEngineFactory extends CFMLEngineFactorySupport {
 				setEngine(engine);
 			}
 			else {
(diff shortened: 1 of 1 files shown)


michaeloffner:
manually adapted the patch
https://github.com/lucee/Lucee/commit/82dd06258762cfbdb898e58dc9381e2cda6cf7bd

PR #2780, opened 2026-04-27
LDEV-6290 refactor scope logging levels

zspitzer:
https://luceeserver.atlassian.net/browse/LDEV-6290

## Summary

Adds the diagnostics needed to actually see when Lucee silently overwrites a live JSession (the bug behind the "intermittent logout" reports going back years). Previously the overwrite happened via a quiet `Map.put` plus an INFO log that read "use existing JSession" even when one had just been created.

Changes in `ScopeContext.java`:

- `createNewJSession` now classifies the prior entry: DEBUG when none, INFO with `lastAccess` + `expired` when replacing a prior JSession, WARN when replacing a non-JSession. Includes httpSessionId, httpSessionAge, maxInactiveInterval for correlation.
- New INFO at the LDEV-2973 trigger site reporting the Lucee-vs-Tomcat clock gap that drove the replacement.
- New DEBUG on `invalidateUserScope` showing sessionType + migrate.
- New `trace()` helper for per-request lifecycle.
- Fixed the misleading "use existing JSession" log so it only fires on the existing-scope path, not after createNewJSession.
- Scope-expiry summary INFO from the periodic Controler tick: `Scope expiry: ended N session(s), N client(s); moved to backing storage N session(s), N client(s)` (only when non-zero).

Storage subsystem (`IKHandlerCache`, `IKHandlerDatasource`, `StorageScopeCookie/File/Memory`, `db/Ansi92`) — demoted noisy per-request INFO lines to DEBUG to make the new INFOs visible at default log levels.

## Test plan

- [x] Full repro suite in `D:/testbeds/jee-session-timeout` (10 repros, fresh instance)
- [x] H1 fires once for the LDEV-2973 trigger (gap=3ms recorded)
- [x] H3 fires 7x across suite — 6x `expired=false` (Railo trigger), 1x `expired=true` (LDEV-2973)
- [x] Scope-expiry summary fires from periodic Controler tick (every-minute and `doit` ticks)
- [x] No new errors / exceptions in `exception.log`

### DIFF
--- a/core/src/main/java/lucee/runtime/type/scope/ScopeContext.java
+++ b/core/src/main/java/lucee/runtime/type/scope/ScopeContext.java
@@ -123,6 +123,10 @@ public final class ScopeContext {
 		error(getLog(), t);
 	}
 
+	public static void trace(Log log, String msg) {
+		if (LogUtil.doesTrace(log)) log.log(Log.LEVEL_TRACE, "scope-context", msg + "; " + ExceptionUtil.getTagContextLine(null));
+	}
+
 	public static void debug(Log log, String msg) {
 		if (LogUtil.doesDebug(log)) log.log(Log.LEVEL_DEBUG, "scope-context", msg + "; " + ExceptionUtil.getTagContextLine(null));
 	}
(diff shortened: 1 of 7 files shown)


michaeloffner:
Merged manually into 7.1 — the branch had diverged from 6.2 and was not cleanly mergeable, so the changes were cherry-picked onto current 7.1 instead.

https://github.com/lucee/Lucee/commit/4a3279814fc8bec289d270e54bba310e8fd5d571

Lösungsversion: 7.1.0.164

PR #2806, opened 2026-08-19
LDEV-4867 / LDEV-4866 preserve SQL comments in queryExecute() and fix query parser bounds (v7 port)

jimpriest:
https://luceeserver.atlassian.net/browse/LDEV-4867

**Description**:
Port of PR #2382 to the 7.0 branch.

queryExecute() strips SQL comments before sending the query to the database, breaking tools that rely on comment directives in executed SQL — notably ProxySQL query routing (https://www.proxysql.com/features/query-routing).

Two bugs in QueryParamConverter.convert():
- LDEV-4867 — /* block */ and -- line comments are silently dropped from the SQL sent to the database
- LDEV-4866 — missing bounds check on charAt(i + 1) causes StringIndexOutOfBoundsException when SQL ends with a bare -, /, or a trailing -- comment with no newline

**Fix**: 
Comments are now preserved in the SQL sent to real datasources. The QoQ internal parser cannot handle comments, so stripSqlComments() is applied only in that path via a new ParserString(String, boolean) constructor.

**Tests**: 
- LDEV4866 5/5, LDEV4867 8/8 passed (49 whitespace variations per statement, tested against both H2 and QoQ). 
- Validated in production with ProxySQL — comment directives now pass through correctly.

**Related PRs**:
Thanks to @bdw429s  who fixed this for v6 (but was never merged)!!

- https://github.com/lucee/Lucee/pull/2382

### DIFF
--- a/core/src/main/java/lucee/commons/lang/ParserString.java
+++ b/core/src/main/java/lucee/commons/lang/ParserString.java
@@ -57,7 +57,18 @@ public final class ParserString {
 	 * @param text CFML Code
 	 */
 	public ParserString(String text) {
-		init(text);
+		init(text, false);
+	}
+
+	/**
+	 * Constructor that optionally strips SQL comments before parsing.
+	 * Use for QoQ SQL parsing where the internal parser cannot handle comments.
+	 *
+	 * @param text SQL text
+	 * @param doIgnoreComments if true, strip SQL comments from text before parsing
+	 */
+	public ParserString(String text, boolean doIgnoreComments) {
+		init(text, doIgnoreComments);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


michaeloffner:
Closing without merging. During review we found a regression in the QoQ comment-stripping used here: it isn't string-literal aware, so `--` / `/* */` markers **inside a quoted string literal** (e.g. `'a--b'`) get stripped and corrupt the query. It also duplicated the comment-scanning logic across three code paths.

We implemented an alternative on 7.0 that routes all comment handling through a single, quote-aware scanner (`SQLComments`) — preserving comments in the SQL sent to the datasource (never counting `?`/`:` inside them as params) while stripping them for the internal QoQ parsers, with literals kept intact. This also fixes the LDEV-4866 bounds issue, and adds tests for the literal-with-comment case.

Fixed in 78e7dc7e3553814a9f7cfedff84144b97bea8cbb. Thanks @jimpriest and @bdw429s for the original work — the approach and tests here build directly on it.

